PR #10, opened 2016-04-26
docs(doctest): Test documentation with markdown-doctest

Widdershin:
Fixes several typos/small issues in the documentation as well.

markdown-doctest will now run after mocha as part of npm test. We should also probably get travis set up so that we quickly notice when things break.

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,7 +135,7 @@ var listener = {
   error: (err) => {
     console.error('The Stream gave me an error: ', err);
   },
-  completed: () => {
+  complete: () => {
     console.log('The Stream told me it is done.');
   },
 }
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -135,7 +135,7 @@ var listener = {
   error: (err) => {
     console.error('The Stream gave me an error: ', err);
   },
-  completed: () => {
+  complete: () => {
```
staltz: Thanks for the typo fixes @Widdershin but could you do these changes instead in the corresponding `markdown/foo` file? Check the first line of this README

staltz: Oh, wait, is this README *already* the result of compiling those files? in that case, ignore...

Widdershin: Oops, missed that! I've actually done it in both locations so everything should be fine.

staltz:
Looks small and good. Thanks!

> We should also probably get travis set up so that we quickly notice when things break.

Yes we should.

Issue #1, opened 2016-04-23
Requires complete listener


elniffi:
So just using this very basic example below to start messing around with xstream in Node.JS it looks like the code assumes there is a "complete" event handler. If I remove it it actually throws an exception when it triggers because it's trying to a run a function that does not exist.

```
'use strict';

var xs = require('xstream').default;

var stream = xs.periodic(50)
    .filter(i => i % 2 === 0)
    .map(i => i * 2)
    .endWhen(xs.periodic(5000).take(1));

stream.addListener({
    next: i => console.log(i)
});

```

```
/Users/elniffi/Documents/Code/xstream-testing/node_modules/xstream/lib/core.js:874
            a[0]._c();
                 ^

TypeError: a[0]._c is not a function
    at Stream._c (/Users/elniffi/Documents/Code/xstream-testing/node_modules/xstream/lib/core.js:874:18)
    at EndWhenOperator.end (/Users/elniffi/Documents/Code/xstream-testing/node_modules/xstream/lib/core.js:308:18)
    at OtherListener._n (/Users/elniffi/Documents/Code/xstream-testing/node_modules/xstream/lib/core.js:278:17)
    at Stream._n (/Users/elniffi/Documents/Code/xstream-testing/node_modules/xstream/lib/core.js:849:18)
    at TakeOperator._n (/Users/elniffi/Documents/Code/xstream-testing/node_modules/xstream/lib/core.js:789:17)
    at Stream._n (/Users/elniffi/Documents/Code/xstream-testing/node_modules/xstream/lib/core.js:849:18)
    at Timeout.intervalHandler [as _repeat] (/Users/elniffi/Documents/Code/xstream-testing/node_modules/xstream/lib/core.js:196:45)
    at Timeout.wrapper [as _onTimeout] (timers.js:408:11)
    at tryOnTimeout (timers.js:224:11)
    at Timer.listOnTimeout (timers.js:198:5)
```
Using Node.js 5.10.1 and Xstream 1.0.2

staltz:
Hi, thank you for taking your time to send us this issue. We appreciate it, however, this is by design. It's required, for two reasons:

- There should not be many `addListener` uses in your code. xstream is not meant for that, it assumes you have a few uses (hence manageable amount) of that
- Having `addListener` polymorphic (types `addListener(next: Function)` and `addListener(listener)`) would be a bit of a hit on performance, and we try to keep it fast.

[...]

PR #12, opened 2016-04-26
feat(addListener): throw an error if next, error or complete functions are missing

Widdershin:
This class of error is a compiler error in TypeScript, but is a runtime error in JS. Currently we get an error like "this._n is not a function" which is quite cryptic to an end user.

I would add a test for this, but since it's a compiler error in TS I can't see any way to do so.

### DIFF
--- a/src/core.ts
+++ b/src/core.ts
@@ -1037,6 +1037,18 @@ export class Stream<T> implements InternalListener<T> {
    * @param {Listener<T>} listener
    */
   addListener(listener: Listener<T>): void {
+    if (typeof listener.next !== 'function') {
+      throw new Error('addListener requires a next function.');
+    }
+
+    if (typeof listener.error !== 'function') {
+      throw new Error('addListener requires an error function.');
+    }
+
+    if (typeof listener.complete !== 'function') {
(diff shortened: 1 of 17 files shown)


staltz:
> I would add a test for this, but since it's a compiler error in TS I can't see any way to do so.

Here's a dirty trick I sometimes use:

```typescript
var listener = <Listener<T>> <any> incompleteListener;
```

:dancers:

erkarl:
Does this mean that the end-user won't be able to subscribe to a stream without providing ***all*** 3 functions (onNext, onError, onCompleted)?

Widdershin:
That is correct, and is in fact by design. See #1.

erkarl:
@Widdershin thanks for clearing it up.

Widdershin:
@staltz fixed the failures (which were the result of done.fail being undefined) and added some tests for these errors. I feel like this should be good to merge 😺

erkarl:
👍

staltz:
Merged as of https://github.com/staltz/xstream/commit/10721b692f5474a88883bddad3f4f9242f05cd43, thanks @Widdershin. I ended up trimming down the code, for size.

PR #14, opened 2016-04-26
Improve xstream performance

milankinen:
It might be possible to achieve 3-10x performance improvements while preserving the current `xstream` semantics and behaviour by tweeking the internal subscription system a little bit.

This PR is about to be a placeholder for those performance optimizations. I'm trying to make the implementations in small independent increments (per commit) so that the optimizations and the motivations behind them are understandable. At least they could give some hints how to improve the performance even if this PR won't be merged into the master.

If you find this PR inappropriate or non-needed, please close it.

Discussion about these performance improvements and their internal implementation can be put into this PR as comments. This is my first time writing TypeScript so I'd also appreciate some review regarding to the code style and its best practices.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "premocha": "npm run lib",
     "mocha": "mocha tests/*.ts tests/**/*.ts --require ts-node/register",
     "postmocha": "rm -rf tests/**/*.js",
-    "test": "npm run lint && npm run mocha && npm run doctest",
+    "test": "npm run lint && npm run mocha",
     "prelib": "rm -rf lib/ && mkdirp lib/ && typings install",
     "lib": "tsc",
     "page-content": "npm run lib && node tools/make-toc.js && node tools/make-factories.js && node tools/make-methods.js && cat markdown/header.md markdown/generated-toc.md markdown/overview.md markdown/generated-factories.md markdown/generated-methods.md markdown/footer.md > .ignore/content.md",
--- a/.gitignore
+++ b/.gitignore
@@ -43,3 +43,7 @@ typings/
 
 # Editors
 .idea/
+
+# Generated dev symbol mappings
+src/**/*.js.map
+tests/**/*.js.map
--- a/perf/combine.js
+++ b/perf/combine.js
@@ -1,0 +1,72 @@
+var Benchmark = require('benchmark');
+var xs = require('../lib/index').default;
+var most = require('most');
+var rx = require('rx');
+var rxjs = require('@reactivex/rxjs')
+var kefir = require('kefir');
+var bacon = require('baconjs');
+var lodash = require('lodash');
+var highland = require('highland');
+
+var runners = require('./runners');
+var kefirFromArray = runners.kefirFromArray;
+
+// Create a stream from an Array of n integers
+// filter out odds, map remaining evens by adding 1, then reduce by summing
+var n = runners.getIntArg(500000);
+var a = new Array(n);
+for(var i = 0; i< a.length; ++i) {
+  a[i] = i;
+}
+
+var suite = Benchmark.Suite('combine(add3) -> filter ' + n + ' x 3 integers');
+var options = {
+  defer: true,
+  onError: function(e) {
+    e.currentTarget.failure = e.error;
+  }
+};
+
+function add3(a, b, c) {
+  return a + b + c;
+}
+
+var xs1 = xs.fromArray(a);
+var xs2 = xs.fromArray(a);
+var xs3 = xs.fromArray(a);
+
+var m1 = most.from(a);
+var m2 = most.from(a);
+var m3 = most.from(a);
+
+var rx1 = rxjs.Observable.from(a);
+var rx2 = rxjs.Observable.from(a);
+var rx3 = rxjs.Observable.from(a);
+
+suite
+  .add('xstream', function(deferred) {
+    runners.runXStream(deferred,
+      xs.combine(add3, xs1, xs2, xs3).filter(even));
+  }, options)
+  .add('most', function(deferred) {
+    runners.runMost(deferred, 
+      most.combineArray(add3, [m1, m2, m3]).filter(even).drain());
+  }, options)
+  .add('rx 5', function(deferred) {
+    runners.runRx5(deferred,
+      rxjs.Observable.combineLatest(rx1, rx2, rx3, add3).filter(even));
+  }, options);
+
+runners.runSuite(suite);
+
+function add1(x) {
+  return x + 1;
+}
+
+function even(x) {
+  return x % 2 === 0;
+}
+
+function sum(x, y) {
+  return x + y;
+}
--- a/perf/multicast.js
+++ b/perf/multicast.js
@@ -44,25 +44,25 @@ function xlater(s) {
 suite
   .add('xstream', function(deferred) {
     var xs1 = xlater(xs.fromArray(a));
-    var xs2 = xs1;
-    var xs3 = xs1;
-    var xs4 = xs1;
+    var xs2 = xs1.map(x => x + 1);
+    var xs3 = xs1.map(x => x + 2);
+    var xs4 = xs1.map(x => x + 3);
     runners.runXStream(deferred,
       xs.combine(addAll, xs1, xs2, xs3, xs4));
   }, options)
   .add('most', function(deferred) {
     var m1 = most.just(most.from(a)).delay(0).switch().multicast();
-    var m2 = m1;
-    var m3 = m1;
-    var m4 = m1;
+    var m2 = m1.map(x => x + 1);
+    var m3 = m1.map(x => x + 2);
+    var m4 = m1.map(x => x + 3);
     runners.runMost(deferred, 
       most.combineArray(addAll, [m1, m2, m3, m4]).drain());
   }, options)
   .add('rx 5', function(deferred) {
     var rx1 = rxjs.Observable.of(rxjs.Observable.from(a)).delay(0).switchMap(x => x).share();
-    var rx2 = rx1;
-    var rx3 = rx1;
-    var rx4 = rx1;
+    var rx2 = rx1.map(x => x + 1);
+    var rx3 = rx1.map(x => x + 2);
+    var rx4 = rx1.map(x => x + 3);
     runners.runRx5(deferred,
       rxjs.Observable.combineLatest(rx1, rx2, rx3, rx4, addAll));
   }, options);
--- a/perf/multicast.js
+++ b/perf/multicast.js
@@ -1,0 +1,86 @@
+var Benchmark = require('benchmark');
+var xs = require('../lib/index').default;
+var most = require('most');
+var rx = require('rx');
+var rxjs = require('@reactivex/rxjs')
+var kefir = require('kefir');
+var bacon = require('baconjs');
+var lodash = require('lodash');
+var highland = require('highland');
+
+var runners = require('./runners');
+var kefirFromArray = runners.kefirFromArray;
+
+// Create a stream from an Array of n integers
+// filter out odds, map remaining evens by adding 1, then reduce by summing
+var n = runners.getIntArg(100000);
+var a = new Array(n);
+for(var i = 0; i< a.length; ++i) {
+  a[i] = i;
+}
+
+var suite = Benchmark.Suite('multicast -> combine(addAll)' + n + ' x 4 integers');
+var options = {
+  defer: true,
+  onError: function(e) {
+    e.currentTarget.failure = e.error;
+  }
+};
+
+
+function xlater(s) {
+  return xs.create({
+    start: p => {
+      setTimeout(() => { 
+        p.next(s);
+        p.complete();
+      }, 0);
+    },
+    stop: () => { }
+  }).flatten();
+}
+
+
+suite
+  .add('xstream', function(deferred) {
+    var xs1 = xlater(xs.fromArray(a));
+    var xs2 = xs1;
+    var xs3 = xs1;
+    var xs4 = xs1;
+    runners.runXStream(deferred,
+      xs.combine(addAll, xs1, xs2, xs3, xs4));
+  }, options)
+  .add('most', function(deferred) {
+    var m1 = most.just(most.from(a)).delay(0).switch().multicast();
+    var m2 = m1;
+    var m3 = m1;
+    var m4 = m1;
+    runners.runMost(deferred, 
+      most.combineArray(addAll, [m1, m2, m3, m4]).drain());
+  }, options)
+  .add('rx 5', function(deferred) {
+    var rx1 = rxjs.Observable.of(rxjs.Observable.from(a)).delay(0).switchMap(x => x).share();
+    var rx2 = rx1;
+    var rx3 = rx1;
+    var rx4 = rx1;
+    runners.runRx5(deferred,
+      rxjs.Observable.combineLatest(rx1, rx2, rx3, rx4, addAll));
+  }, options);
+
+runners.runSuite(suite);
+
+function add1(x) {
+  return x + 1;
+}
+
+function even(x) {
+  return x % 2 === 0;
+}
+
+function sum(x, y) {
+  return x + y;
+}
+
+function addAll(a, b, c, d) {
+  return a + b + c + d;
+}
--- a/perf/runners.js
+++ b/perf/runners.js
@@ -104,7 +104,7 @@ function runXStream(deferred, xstream) {
   xstream.addListener({
     next: noop,
     complete: function() {
-      deferred.resolve();
+      setImmediate(() => deferred.resolve());
     },
     error: function(e) {
       deferred.benchmark.emit({ type: 'error', error: e });
--- a/src/core.ts
+++ b/src/core.ts
@@ -1167,9 +1167,12 @@ export class Stream<T> {
    * @param {Listener<T>} listener
    */
   addListener(listener: Listener<T>): void {
-    (<InternalListener<T>> (<any> listener))._n = listener.next;
-    (<InternalListener<T>> (<any> listener))._e = listener.error;
-    (<InternalListener<T>> (<any> listener))._c = listener.complete;
+    const s: Subscription<T> = {
+      listener,
+      observer: new Observer(this.source, listener)
+    };
+    this._subs.push(s);
+    s.observer.start();
   }
 
   /**
@@ -1178,6 +1181,26 @@ export class Stream<T> {
    * @param {Listener<T>} listener
    */
   removeListener(listener: Listener<T>): void {
+    for (let i = 0; i < this._subs.length; i++) {
+      const s = this._subs[i];
+      if (s.listener === listener) {
+        s.observer.stop();
+        this._subs.splice(i, 1);
+        return;
+      }
+    }
+  }
+  
+  _n(t: T): void {
+  }
+
+  _e(err: any): void {
+  }
+
+  _c(): void {
+  }
+
+  _x(): void {
   }
 
   _add(il: InternalListener<T>): void {
--- a/src/core.ts
+++ b/src/core.ts
@@ -71,6 +71,12 @@ export interface Listener<T> {
 
 
 
+// indexed value type
+interface Indexed<T> {
+  i: number;
+  x: T;
+}
+
 // class for easy deferred task creation and calcelation
 class Deferred { 
   private t: any
@@ -412,6 +418,78 @@ export interface CombineInstanceSignature<T> {
     stream5: Stream<T5>): Stream<R>;
   <R>(project: (...args: Array<any>) => R, ...streams: Array<Stream<any>>): Stream<R>;
 }
+
+
+class CombineSource implements Source<Indexed<any>> {
+  private sinks: Array<IdxSink<any>>;
+  constructor(private sources: Array<Source<any>>) {
+    this.sinks = null;
+  }
+  start<B>(sink: Sink<Indexed<any>, any>): void {
+    let sources = this.sources, n = sources.length, idxs: IdxSink<any> = null;
+    this.sinks = [];
+    for (let i = 0; i < n && this.sinks; i++) {
+      idxs = new IdxSink(sink, i);
+      this.sinks.push(idxs);
+      sources[i].start(idxs);
+    }
+  }
+  stop<B>(sink: Sink<Indexed<any>, B>): void {
+    const s = this.sinks, n = s.length, sources = this.sources;
+    this.sinks = null;
+    for (let i = 0; i < n; i++) sources[i].stop(s[i]);
+  }
+}
+
+class Combine<A> extends Combinator<Indexed<any>, A> {
+  private n: number;
+  constructor(sources: Array<Source<any>>, private fn: (...args: Array<any>) => A) {
+    super(new CombineSource(sources));
+    this.n = sources.length;
+  }
+  run(sink: Sink<A, any>) {
+    return new CombineSink(sink, this.fn, this.n);
+  }
+}
+
+
+class IdxSink<A> extends MCastSink<A, Indexed<A>> {
+  constructor(sink: Sink<Indexed<A>, any>, private i: number) {
+    super(sink);
+  }
+  next(x: A) {
+    this.s.next({i: this.i, x});
+  }
+}
+
+
+class CombineSink<A> extends MCastSink<Indexed<any>, A> {
+  vals: Array<any>;
+  nEnd: number;       // number of streams that still needs to end
+  nVal: number;       // number of streams still waiting for initial value
+  constructor(sink: Sink<A, any>, private fn: (...args: Array<any>) => A, n: number) {
+    super(sink);
+    const vals = this.vals = Array(n);
+    this.nEnd = this.nVal = n;
+    for (let i = 0; i < n; i++) vals[i] = none;
+  }
+  next(x: Indexed<A>) {
+    const i = x.i, val = this.vals[i], f = this.fn;
+    const left = (val === none && --this.nVal) || this.nVal;
+    this.vals[i] = x.x;
+    left === 0 && this.s.next(invokeF(f, this.vals));
+  }
+  end(): void {
+    if (--this.nEnd === 0) {
+      this.s.end();
+    }
+  }
+  cleanup() {
+    this.vals = null;
+    super.cleanup();
+  }
+}
+
 
 class CombineListener<T> implements InternalListener<T> {
   constructor(private i: number,
@@ -1366,8 +1444,10 @@ declare type Subscription <T> = {
 
 export class Stream<T> {
   private _subs: Array<Subscription<T>>;
-  constructor(public source: Combinator<any, T>) {
+  public source: Combinator<any, T>;
+  constructor(src: Combinator<any, T>) {
     this._subs = [];
+    this.source = src;
   }
  
   /**
@@ -1643,7 +1723,7 @@ export class Stream<T> {
   static combine: CombineFactorySignature =
     function combine<R>(project: CombineProjectFunction,
                         ...streams: Array<Stream<any>>): Stream<R> {
-      throw new Error("Not implemented yet");
+      return new Stream(new Combine<R>(streams.map(s => s.source), project));
     };
 
   /**
@@ -1984,7 +2064,7 @@ export class Stream<T> {
   combine: CombineInstanceSignature<T> =
     function combine<R>(project: CombineProjectFunction,
                         ...streams: Array<Stream<any>>): Stream<R> {
-      throw new Error("Not implemented yet");
+      return Stream.combine(project, this, ...streams);                    
     };
 
   /**
--- a/src/core.ts
+++ b/src/core.ts
@@ -522,6 +522,34 @@ class PeriodicProducer implements Source<number> {
   }
 }
 
+class CustomProducer<A> implements Source<A> {
+  private lis: Listener<A>;
+  private sink: Sink<A, any>;
+  constructor(private prod: Producer<A>) { 
+    this.sink = null;
+    this.lis = {
+      next: x => {
+        const s = this.sink;
+        try {
+          s && s.next(x);
+        } catch (err) {
+          handleError(err, s);
+        }
+      },
+      complete: () => this.sink && this.sink.end(none),
+      error: err => this.sink && this.sink.end(err)
+    };
+  }
+  start<B>(sink: Sink<A, B>): void {
+    this.sink = sink;
+    this.prod.start(this.lis);
+  }
+  stop<B>(sink: Sink<A, B>): void {
+    this.sink = null;
+    this.prod.stop();
+  }
+}
+
 export class DebugOperator<T> implements Operator<T, T> {
   private out: Stream<T> = null;
 
@@ -1223,7 +1251,7 @@ export class Stream<T> {
    * @return {Stream}
    */
   static create<T>(producer?: Producer<T>): Stream<T> {
-    return new Stream(null);
+    return !producer ? Stream.empty() : new Stream(ident(new CustomProducer(producer)));
   }
 
   /**
@@ -1252,7 +1280,7 @@ export class Stream<T> {
    * @return {Stream}
    */
   static never(): Stream<any> {
-    throw new Error("Not implemented yet");
+    return Stream.create({start: noop, stop: noop});
   }
 
   /**
@@ -1270,7 +1298,10 @@ export class Stream<T> {
    * @return {Stream}
    */
   static empty(): Stream<any> {
-    return new Stream(null);
+    return Stream.create({
+      start: lis => lis.complete(),
+      stop: noop
+    });
   }
 
   /**
@@ -1290,7 +1321,10 @@ export class Stream<T> {
    * @return {Stream}
    */
   static throw(error: any): Stream<any> {
-    throw new Error("Not implemented yet");
+    return Stream.create({
+      start: lis => lis.error(error),
+      stop: noop
+    });
   }
 
   /**
--- a/src/core.ts
+++ b/src/core.ts
@@ -805,42 +805,31 @@ export class FoldOperator<T, R> implements Operator<T, R> {
   }
 }
 
-export class LastOperator<T> implements Operator<T, T> {
-  private out: Stream<T> = null;
-  private has: boolean = false;
-  private val: T = <T> empty;
-
-  constructor(public ins: Stream<T>) {
+class Last<A, B> extends Combinator<A, A> {
+  constructor(source: Source<A>) {
+    super(source);
   }
-
-  _start(out: Stream<T>): void {
-    this.out = out;
-    this.ins._add(this);
+  run(next: Sink<A, any>) {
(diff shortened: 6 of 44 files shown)


package.json
```diff
@@ -10,7 +10,7 @@
     "premocha": "npm run lib",
     "mocha": "mocha tests/*.ts tests/**/*.ts --require ts-node/register",
     "postmocha": "rm -rf tests/**/*.js",
-    "test": "npm run lint && npm run mocha && npm run doctest",
+    "test": "npm run lint && npm run mocha",
```
Widdershin: Curious, did you mean to include the removal of running doctest as part of this PR, or was it just a local convenience that got committed?

Widdershin: Wait nevermind, I see now that a bunch of stuff isn't implemented, that makes sense.

milankinen:
Okay, the first iteration is pushed to the `perf` branch now.

* Interfaces and basic pipeline
* `take` `last` `filter` `map` and `fold` implementation
* `fromArray` and `periodic` producers 

## The idea

The idea behind the new subscription system is to treat subscription as a *function composition* - subscription from producer to listener is basically just one function call. When new event is produced, this "function" is called and the outer function of the composed function is the actual event handler (`listener.next`). 

The idea is pretty much same as in the current `xstream` implementation but having less computation steps. This is achieved by using `Sink`s, `Source`s and `Combinator`s. 

When observable chain is created, each step in that chain contain a `Combinator` that extends `Source`. Each combinator knows its source (which is again a combinator). Combinator is almost like an operator but it's not exactly same: **combinator does not do any computation**, instead it encapsulates this computation.

When new listener is added, stream's combinator is invoked and the next computation step is passed as an argument to the combinator. Combinator now just *combines* its computation to the listener and passes it down to its source (which again combines the encapsulated computation so on...)

Computation steps are modeled as `Sink`s. Sinks are basically just functors `f: x -> y`, but `fmap` renamed to `next`. `Sink<A, B>` constructor is `constuctor :: Sink<B, any>`. 

Finally `Source` receives the sink. And because that sink is only a composition of sinks the `sink.next` is basically jus a function: `next = x => lastSink.next(foldSink.next(mapSink.next(filterSink.next(x))))` In contrast, in the current `xstream` implementation those function calls "bounce" from operators to streams to operators and so one. The call chain is more complex and the function calls are bigger so it prevents some compiler inline optimizations. 

### Where to put the checks?

The assumption is that number of next events is much bigger than number of error/complete events, which means that sinks must be as small and fast as possible. This means that they can't have any checks. Instead, those checks must be placed into the beginning (producer) or end (listener) of the computation chain.

Fortunately we have 100% control over the chain from producer to listener. That's why we need yet another concept: `Observer`. Observer sits between sinks and the listener and it is responsible for correct subscription cleanup and sync/async handling. 

When subscription ends, observer is responsible for invoking `stop` from it's source which (recursively) leads that the whole subscription gets disposed. Producers must also be aware of those disposals and ensure that they send *only one* end/error event and no next events after that.

## Results

With new implementation:
```
perf (perf) $ node filter-map-reduce.js 
filter -> map -> reduce 1000000 integers
-----------------------------------------------
xstream     241.87 op/s ±  1.21%   (41 samples)
most        333.35 op/s ±  1.47%   (78 samples)
^C
perf (perf) $ node filter-map-fusion.js 
filter -> map -> reduce 1000000 integers
-----------------------------------------------
xstream      55.99 op/s ±  2.02%   (35 samples)
most         75.63 op/s ±  6.11%   (52 samples)
^C
```

With current:
```
perf (master) $ node filter-map-fusion.js 
filter -> map -> reduce 1000000 integers
-----------------------------------------------
xstream      18.22 op/s ±  2.80%   (26 samples)
^C
perf (master) $ node filter-map-fusion.js 
filter -> map -> reduce 1000000 integers
-----------------------------------------------
xstream      18.92 op/s ±  3.12%   (29 samples)
^C
```

And note that those performance improvements don't include sink fusion part yet, which would improve the performance even further.

## Some notes

Sharp reader may notice that the implementation is unicast instead of multicast and doesn't support any error handling yet. I'll add them next... 😉

briancavalier:
Good to see many of [most.js's architectural patterns](https://github.com/cujojs/most/wiki/Architecture) being implemented here, like synchronous call stack based propagation (instead of trampolines), fusion, etc.

milankinen:
Yeah. The new implementation is still almost a clone from `most`. The multicasting and error handling part is gonna change it a little bit.

milankinen:
## Step 2 - error handling

The current implementation of `xstream` places try-catch blocks around every operator (like `.map`, `.filter` etc...). This is very expensive and prevent some compiler optimizations.

However, we know that the following applies in the producers: 
```
next = x => lastSink.next(foldSink.next(mapSink.next(filterSink.next(x))))
``` 
The solution is simple: just wrap the producer block with `try - catch`!

Now that we have our error caught, we have to do something about it. If we think what parts of the application are interested in errors, the answer is: listeners (=observers). Now we have to just get the errors to those observers. But how? In producer we know only `sink`. 

But observer is sink too, right? Let's define `Sink.prototype.handers(): Array<Sink<any, any>>`. The default implementation is just `return this.s.handlers()` but Observers must override this so that they actually return `[ this ]`. Now those handlers recursively end up to the producer, which can use them  `handlers.forEach(h => h.end(err))` (simplified).

### `replaceError`

Replace error is not an exception here. It just overrides the `handlers` so that it doesn't pass the exceptions to the upstream but handles them by itself instead. 

The interesting part of `.replaceError` is that it should (at least if I remember right our discussion with @staltz) be able to restart the source stream if used like `stream.replaceError(() => stream)`.  In multicast case we can't stop the producer and start it synchronously due to ref-counting: we would just re-subscribe to the hot (ended) stream and get nothing (I think there might be a bug in the current implementation as well, I might create another issue for that). 

That's why we must *stop the source immediately but restart it asynchronously* so that other listeners can stop the errored source during the current tick, thus making the source cold again. During the next tick, the source can be re-started again correctly. 

### Perf?

Again, we get pretty good results:
```
perf (perf *) $ node filter-map-reduce.js 
filter -> map -> reduce 1000000 integers
-----------------------------------------------
xstream     213.24 op/s ±  1.90%   (42 samples)
most        327.71 op/s ±  1.93%   (76 samples)
^C
perf (perf *) $ node filter-map-fusion.js 
filter -> map -> reduce 1000000 integers
-----------------------------------------------
xstream      44.79 op/s ±  5.27%   (37 samples)
most         73.26 op/s ±  5.35%   (50 samples)
^C
```

### Multicasting

Smart reader may notice that this works only for unicast streams because we can't know in which branch the exception was thrown, if the stream is multicasted. Just wait for the next step!

briancavalier:
Forgive me for jumping in again, @staltz.

@milankinen I'm curious to hear your intentions with this PR.

As you've mentioned, there are striking similarities with quite a few things in most.js, including the [isolation of try/catch near producers](https://github.com/cujojs/most/wiki/Architecture#error-propagation).  I'm glad it has been an inspiration.  However, this PR does give the impression that your goal may simply be to duplicate a subset of most.js in typescript.

Perhaps the three of us should discuss the how we might work together to achieve some common goals?

juhovh:
This PR is useful as a proof-of-concept, but too large to merge as is and contains also some internal architectural changes. So there are some things to be considered with performance improvements:

1) do they also improve code readability and/or productivity?
2) is the related code currently a bottleneck in some specific real-world usecase?
3) do they have to be done immediately in order to not break the external APIs?

To be honest I'm not sure if this PR answers "yes" to any of those questions. But if it does, then discussion should be started about how to proceed. In general it's better to avoid premature optimizations before the actual goals are clear enough.

milankinen:
@briancavalier 

I'm very sorry if the implementations overlap. I haven't looked into `most` source codes in detail - the drive for this PR came from the last weekend's "hackathon" when `xstream` was announced. I've played with `most` and know it's extremely fast so when the `xstream` was announced, I made a personal challenge to see how much I can improve its performance while maintaining the exact same semantics (sync subscription, async disposal, multicast, error handling).

Yes, I took a sneek peek into `most`'s source code to figure out why it's so damn fast. I immediately saw the source-sink pattern and realized that it might be the key for the amazing performance. After that I've been on my own, although it is nice to notice that I've ended up using mostly (pun not intended) same patterns as you guys are using in e.g. error handling (though there seem to be some differences, e.g. `most` sends errors through separate error channel whereas new `xstream` first gathers the observers and then send error directly to them). After some brainstorming me and my colleague also figured out a way to make ad-hoc multicasting without sacrificing the performance (and that's my next step btw).

So my intentions: I've not done JS engine optimizations before and this PR is kind of like some "trial'n'error" way for me to understand it (for example today I learned that comment blocks inside your functions prevent inlining 😧). I've also received some hints from my colleagues. 😄 I'd truly like to create (or at least prove to myself that I can) a fast observable implementation while preserving the `xstream`'s "beginner-friendly" behaviour. I'm really sorry if this (unintentionally) leads to some duplicate concepts with `most`. 😞 

I'm very interested in listening to what you have in your mind.

staltz:
Thanks both @briancavalier and @juhovh.

To start with, here are the goals of xstream, from most important to least:

- Have an intuitive API with just <30 operators
- Tailored for Cycle.js apps (multicast only, endWhen and imitate)
- Small in kB size, goal is 30kB
- TypeScript
- Be in the league of "fast stream libraries" (benchmark is: at least as fast as RxJS v5)

So, being extremely fast is not a goal in xstream, but if we can improve the performance without sacrificing any of the goals above, great.

On the other hand, @briancavalier is so genius at performance that I'd gladly yield the title of "fastest stream library" to him, I don't need to try to compete or outperform that. That said, I'll keep most.js as an inspiration on how to achieve more performance, as long as it's not a copy.

Hope that clarifies

staltz:
PS: not to mention when we talk about "performance", it's just "microbenchmarks", so no bottlenecks in sight, like @juhovh mentioned.

milankinen:
@juhovh

Thanks, those are good questions!

>1) do they also improve code readability and/or productivity?

I'm not unfortunately the one to answer that question. You can take a look at how to implement new operators (I've tried to make them separate commits). Adding new operator is quite straightforward: add new combinator and sink and create a new public method to `Stream`. This also means that operators are "isolated", hence adding new / modifying old ones is safe and don't break the others. Is it more readable or gives more productivity compared to the current implementation? That I can't say.

>2) is the related code currently a bottleneck in some specific real-world usecase?

Why to stop at something if you can go further? Cycle is architecturally pretty heavy and faster streaming libs / driver implementations have been the proposed solution so far. Why not to build as fast "beginner-friendly" streaming library as we can?

>3) do they have to be done immediately in order to not break the external APIs?

Nope. But as more time passes and features come in, then more time it takes to re-write them by using new implementation.

milankinen:
>Small in kB size, goal is 30kB

And yes. That's also one thing I'm aware of and want to preserve in future. 😄

juhovh:
@milankinen 

Thanks for your clarifications, some extra comments:

> 1)

Yes, in most cases this is not a clear question, in which case I guess it's up to the rest of the team. I didn't mean you should be able to answer all the questions. :)

> 2)

Definitely a great long term goal, but I wanted to bring real world use cases into the discussion just because a 10x optimisation of a 30ms code in a total of 300ms operation makes it still take 273ms. The numbers here are just out of thin air, and since streaming library is the core of Cycle, the benefits are probably much bigger here, but I'm sure you get the point. Still it would be cool to have also results like "rendering this sample page became 50% faster" instead of just running benchmarks.

> 3)

If the goal of xstream is to be a very small library, then the time needed for refactoring shouldn't grow that much, right?

In any case, I think there is a potential case here. I see you have commented out most of the tests, do you think the performance improvements could be done in small increments without breaking any of the existing tests in the process?

staltz:
> do you think the performance improvements could be done in small increments without breaking any of the existing tests in the process?

Good reminder. For instance one thing now is that we don't have perf tests for `combine()`, and it is a critical operator. A PR that would add perf tests for `combine` and also tweak a few minor things with its implementation would be very valuable if someone wants to help making xstream faster.

milankinen:
>a 10x optimisation of a 30ms code in a total of 300ms operation makes it still take 273ms

Yes, good point there. When evaluating the overall performance, there are also many other parts of code having impact on the performance, e.g. vdom implementation. I agree that there are many lower hanging fruits to do the performance optimizations (like the vdom I mentioned) but they're just not so djuizy. 😄 

I also think these optimizations in this way: stream library is the backbone of your application which means it's being used immensely. If you manage to drop 1ms operation to 0.9ms somehow, it gives you 10 sec to do other stuff if that operation is repeated 100k times.

>If the goal of xstream is to be a very small library, then the time needed for refactoring shouldn't grow that much, right?

Qood question. When is the "right time" to make the refactoring then? In my experience, the more mature some app/library gets, the more expensive it becomes to refactor.

>I see you have commented out most of the tests, do you think the performance improvements could be done in small increments without breaking any of the existing tests in the process?

Unfortunately the internal implementation requires so much changes that having a working interoperation between old and new implementation is not a trivial task to do. It actually might be possible but would require a lot of extra work and probably violates the size goal. Perhaps I should investigate that option as well? 

The reason I commented the tests instead of removing them, is that I'm going to uncomment them as I get new operator implementations working, finally having all tests working again, as they were before the optimization.

briancavalier:
@milankinen Thank you for the thoughtful and honest reply.  I'm a big fan of learning by doing as well.

@staltz Thanks for the kind words.  In reality, it's been [99% perspiration](http://www.phrases.org.uk/meanings/genius-is-one-percent-perspiration-ninety-nine-percent-perspiration.html): lots of time spent learning, profiling, trail-and-erroring, building and rebuilding.

Of course, I've been inspired by many others too (rxjs, kefir, reactive-banana, FRP.Reactive, to name only a few).  It has always been my hope that any innovations in most.js feed back to the community.  So, in seeing this PR, my concern was about *duplication* (given all the *perspiration*), rather than about *inspiration*.

It sounds like if this PR is accepted, it could end up taking xstream in a direction that is inspired by most.js.  I'm cool with that :)

milankinen:
@briancavalier I can see you concern. Duplication is definitely not my goal. The call stack based event propagation is a brilliant idea that I (shamelessly) copied from `most`, but after that I've tried to go on my own.

AlexGalays:
I actually need xstream (or kefir's) semantics for my own lib, where a MemoryStream can yield its last value synchronously if it has one. That would be a major difference with most, where everything is async on purpose.

TylorS:
@AlexGalays Everything is not async in most. The first event is always async, to avoid missing events, the second and every event thereafter is 100% synchronous unless you use async operators like `delay()`.

milankinen:
## Step 3 - multicasting

Okay. So far the optimized implementation has been *unicast* whereas `xstream` streams are *multicast* by default. Let's make it so. If you've looked at the code, you may have noticed that each sink contains a reference to the next, e.g. `MapSink`:
```typescript
class MapSink<A, B> extends Sink<A, B> {
  constructor(sink: Sink<B, any>, private fn: (x: A) => B) {
    super(sink);
  }
  next(x: A): void {
    const f = this.fn;
    this.s.next(f(x));     // <<< "this.s" here!!
  }
}
```
What multicasting actually does is that it just delegates those next events to multiple sinks instead of only one. Hence the naive implementation would just be:
```typescript
next(x: A): void {
    const f = this.fn;
    this.sinks.forEach(s => s.next(f(x)));
}
```
But that is not performant at all. How could we improve that. In real apps, most of the streams are unicast and there are only few spots where the computation actually diverges. Think about this observable chain:
```js
const response = textChange
  .map(toText)
  .debounce(300)
  .filter(atLeast3Len)
  .map(toReq)
  .flatten()

const items = response.map(toItems)
const meta = response.map(toMeta)
```
All intermediate steps are unicast and only `response` stream is multicasted. So our goal is definitely to preserve the unicast behaviour (and perf) in all of those places where multicast is not needed.

So let's take a look again the sink implementation and make a little addition:
```diff
class MapSink<A, B> extends Sink<A, B> {
  constructor(sink: Sink<B, any>, private fn: (x: A) => B) {
    super(sink);
  }
+ subscribe(sinks: Array<Sink>) {
+   this.s = sinks.length === 1 ? sinks[0] : new MulticastingSink(sinks)
+ }
  next(x: A): void {
    const f = this.fn;
    this.s.next(f(x));  
  }
}
```
Now we are using the original `sink` (unicast) initially but if there arrives more subscribers, `this.s` gets replaced with `MulticastingSink` that does the for-each loop. Of course we need store all subscriptions somewhere so that we can react to subscriptions/disposes and (re)build the call stack when changes occur. That's why we have the `Combinator` abstraction:
```typescript
export abstract class Combinator<A, B> implements Source<B> {
  source: Source<A>;
  sink: Sink<A, B>;
  sinks: Array<Sink<B, any>>;
  
  constructor(source: Source<A>) {
    this.source = source;
    this.sink = null;
    this.sinks = [];
  }
  
  abstract run(next: Sink<B, any>): Sink<A, B>
  
  start(next: Sink<B, any>): void {
    this.sinks.push(next);
    if (this.sink === null) {
      const s = this.sink = this.run(next);
      this.source.start(s);
    } else {
      const s = this.sink
      s.s = new MulticastingSink(this.sinks.slice());
    }
  }
  
  stop(sink: Sink<B, any>): void {
    const left = remove(this.sinks, sink);
    const s = this.sink;
    if (left === 0) {
      this.sink = null;
      this.source.stop(s);
      s.cleanup();
    } else if (left === 1) {
      s.s.cleanup();
      s.s = this.sinks[0];
    } else {
      s.s.cleanup();
      s.s = new MulticastingSink(this.sinks.slice());
    }
  }
}
``` 
But how can we ensure that our sinks contain `.s` property which can be replaced? Types actually come very handy in this case. Let's just make the contract of our "multicast-supported" sink official:
```typescript
export interface MSink<A, B> extends Sink<A, B> {
  s: Sink<B, any>;
}
```
And now we can ensure that all our combinators actually return sinks that support multicasting:
```diff
export abstract class Combinator<A, B> implements Source<B> {
  source: Source<A>;
- sink: Sink<A, B>;
+ sink: MSink<A, B>;
  sinks: Array<Sink<B, any>>;
  
  constructor(source: Source<A>) {
    this.source = source;
    this.sink = null;
    this.sinks = [];
  }
  
- abstract run(next: Sink<B, any>): Sink<A, B>
+ abstract run(next: Sink<B, any>): MSink<A, B>
  
  // same stuff as before....
}
```
We can also ensure that our streams are all multicasted:
```diff
export class Stream<T> {
  private _subs: Array<Subscription<T>>;
- public source: Source<T>;
+ public source: Combinator<any, T>;
  // ...
}
```
But how about producers? They implement `Source` instead of `Combinator`? Easy. Because we know that call stack based sinks are relatively cheap, we can just create "identity" combinator that deals the multicasting and lets us keep the producers clean and simple:
```typescript
class MulticastedIdentity<A> extends Combinator<A, A> {
  constructor(source: Source<A>) {
    super(source);
  }
  run(next: MSink<A, any>) {
    return new IdentitySink(next);
  }
}
class IdentitySink<A> extends MCastSink<A, A> {
  constructor(sink: Sink<A, any>) {
    super(sink);
  }
  next(x: A) {
    this.s.next(x);
  }
}
```
Now we can just wrap the producers with that `MulticastedIdentity` combinator.

...actually that's not all - there is still error handling to do but I'll write a separate comment for that.

milankinen:
## Step 4 - flatten and combine

My goal is to test multicasting performance by using the following code:
```js
suite
  .add('xstream', function(deferred) {
    var xs1 = xlater(xs.fromArray(a));
    var xs2 = xs1.map(x => x + 1);
    var xs3 = xs1.map(x => x + 2);
    var xs4 = xs1.map(x => x + 3);
    runners.runXStream(deferred,
      xs.combine(addAll, xs1, xs2, xs3, xs4));
  }, options)
  .add('most', function(deferred) {
    var m1 = most.just(most.from(a)).delay(0).switch().multicast();
    var m2 = m1.map(x => x + 1);
    var m3 = m1.map(x => x + 2);
    var m4 = m1.map(x => x + 3);
    runners.runMost(deferred, 
      most.combineArray(addAll, [m1, m2, m3, m4]).drain());
  }, options)
  .add('rx 5', function(deferred) {
    var rx1 = rxjs.Observable.of(rxjs.Observable.from(a)).delay(0).switchMap(x => x).share();
    var rx2 = rx1.map(x => x + 1);
    var rx3 = rx1.map(x => x + 2);
    var rx4 = rx1.map(x => x + 3);
    runners.runRx5(deferred,
      rxjs.Observable.combineLatest(rx1, rx2, rx3, rx4, addAll));
  }, options);

runners.runSuite(suite);

function xlater(s) {
  return xs.create({
    start: p => {
      setTimeout(() => { 
        p.next(s);
        p.complete();
      }, 0);
    },
    stop: () => { }
  }).flatten();
}
```
The idea is to delay the outer stream so that all four listeners/subscribers get added before the actual data starts to flow through the observables, thus it'll be multicasted to all four subscribers (`xs1`, `xs2`, `xs3`, `xs4` and their most/rx counterparts). 

However, before we can execute that test, we must implement `flatten` and `combine`.

### `flatten`

`flatten` (or switch) is the easier one of those two, so let's take it first. The current `xstream` implementation is pretty good and performs quite well:
```
perf (master) $ node switchMap.js 
switchMap 1000 x 1000 streams
-----------------------------------------------
xstream      14.34 op/s ±  1.09%   (66 samples)
most         44.52 op/s ±  4.25%   (56 samples)
rx 5         20.49 op/s ±  1.21%   (28 samples)
```
However, there is still two parts that can be optimized: `try-catch` block and `_remove(listener)` logic. The `try-catch` block is easy: we can just omit it because the `try-catch` block in the inner subscription's producer + error handling deals it. The only thing we need to do is to ensure that when the outer sink is cleaned up, it also stops the inner producer:
```js
class FlattenSink<A> extends MCastSink<Stream<A>, A> {
  // ...
  cleanup() {
    const src = this.src;
    this.src = null;
    src && src.stop(this.s as FInnerSink<A>);
    super.cleanup();
  }
}
```
Listener removal is a bit more complex case but anything special there either. Just add member `d: Deferred` which provides method `.cancel` - when new inner observable arrives to the outer sink, we just cancel the existing (if any) deferred task and create a new one which does the disposal of the previous subscription and makes a new one by using new inner source. If the new inner source is same as the currently used, then we don't need to do anything.
```js
  next(x: Stream<A>): void {
    const next = x.source;
    this.d && this.d.cancel();
    this.d = defer(() => this._next(next));
  }
  _next(next: Source<A>) {
     const src = this.src, same = next === src;
     src && !same && src.stop(this.s as FInnerSink<A>);
     this.src = null;
     (this.src = next) && !same && next.start(this.s as FInnerSink<A>);
  }
```
(Of course the deferred task must be cleaned up as well, but it's trivial)

Let's take a look at the performance results with the new implementation:
```
perf (perf) $ node switchMap.js 
switchMap 1000 x 1000 streams
-----------------------------------------------
xstream      77.74 op/s ±  2.45%   (71 samples)
most         64.57 op/s ±  0.96%   (73 samples)
rx 5         20.17 op/s ±  1.95%   (28 samples)
```
I'm actually pretty stunned that `xstream` can outperform `most`. That makes me wonder if I've done everything correctly... 😕 But at least all `xstream` tests pass!

### `combine` 

The most important thing in `combine` implementation is to achieve `O(1)`. The current implementation is already in the right track: it assigns array indexes to listeners during the listener creation which makes write operations `O(1)`. The performance according to my [combine test](https://github.com/milankinen/xstream/blob/perf/perf/combine.js) is indeed good:
```
perf (master) $ node combine.js 
combine(add3) -> filter 500000 x 3 integers
-----------------------------------------------
xstream       8.03 op/s ±  1.25%   (41 samples)
most          3.92 op/s ±  1.53%   (23 samples)
rx 5          8.85 op/s ±  1.52%   (25 samples)
-----------------------------------------------
```
The thing that bothers me, is `most`'s extremely bad performance. Perhaps my test just don't favor `most` or misuses it somehow so I'm just focusing on the comparison between the current and optimized `xstream` implementation. 

That said, the pain point is the initial check (`CombineProducer.prototype.up`) which is `O(n)` and requires array looping. The same thing could've be achieved by using cheap identity check and internal counter value:
```js
  next(x: Indexed<A>) {
    const i = x.i, val = this.vals[i], f = this.fn;
    const left = !this.nVal ? 0 : val === none ? --this.nVal : this.nVal;
    this.vals[i] = x.x;
    left === 0 && this.s.next(f.apply(null, this.vals));
  }
```
I also noticed that comparing object and number is quite expensive, hence the ternary operator.

My initial implementation used `f.apply(null, this.vals)` which was terribly slow. That made me to find better solutions. Fortunately the current `xstream` implementation provided a pretty good starting point:
```js
function invoke(f: Function, args: Array<any>) {
  switch (args.length) {
  case 0: return f();
  case 1: return f(args[0]);
  case 2: return f(args[0], args[1]);
  case 3: return f(args[0], args[1], args[2]);
  case 4: return f(args[0], args[1], args[2], args[3]);
  case 5: return f(args[0], args[1], args[2], args[3], args[4]);
  default: return f.apply(void 0, args);
  }
```
Performance gain was huge. However, I didn't want to stop there. How about if we get rid of the switch case?
```js
const iTable = [
  (f, args) => f(),
  (f, args) => f(args[0]),
  (f, args) => f(args[0], args[1])
]
function invoke(f, args) {
   const n = args.length;
   return n < N ? iTable[n](f, args) : f.apply(void 0, args);
}
``` 
I also noticed that V8 doesn't make any difference between generated functions and "hard-coded" ones in terms of performance, hence we can easily reduce the size of the library by using `eval`
```js
const iTable = times(N, i => 
  eval(`(function invoke${i}(f, x) { return f(${times(i, i => "x["+i+"]").join(",")}); })`));
// ...
```
We can still go further. We know `n` during the sink creation, right? Let's then choose the right invoke implementation during the sink constructor and store it.
```js
class CombineSink<A> extends MCastSink<Indexed<any>, A> {
  invoke: (f: (...args: Array<any>) => A, vals: Array<any>) => A;   
  // ...
  constructor(sink: Sink<A, any>, private fn: (...args: Array<any>) => A, n: number) {
    super(sink);
    this.invoke = n < 10 ? iTable[n] : invokeN;
    // ...
  }
  next(x: Indexed<A>) {
    const i = x.i, val = this.vals[i], f = this.fn, invoke = this.invoke;
    const left = !this.nVal ? 0 : val === none ? --this.nVal : this.nVal;
    this.vals[i] = x.x;
    left === 0 && this.s.next(invoke(f, this.vals));
  }
  // ...
}
```  

... and here we go! The improved performance results:
```
perf (perf *) $ node combine.js 
combine(add3) -> filter 500000 x 3 integers
-----------------------------------------------
xstream      33.22 op/s ±  2.17%   (75 samples)
most          3.84 op/s ±  0.82%   (23 samples)
rx 5          9.56 op/s ±  1.03%   (26 samples)
-----------------------------------------------
```

milankinen:
## Step 5 - optimize multicasting

As I said before, multicasting to `n` subscriptions requires `n` `try-catch` blocks and those blocks are poison to the performance. How to reduce the number of them without sacrificing the error handling?

My work with the `combine` operator gave me some ideas. Like in `combine` we know the subscription count (`sinks.length`) during the `MulticastingSink` creation. But how can we utilize that information? We've used `iTable` lookup table for the `invoke` functions, so let's try to create similar table for multicast functions:

```js
const mcastTable = [
   mcast0,
   mcast1,
   mcast2,
   ...
]

function mcast0(x, sinks, handleError) {}
function mcast1(x, sinks, handleError) {
  let i = 1
  try {
    sinks[--i].next(x)
  } catch (e) {
    handleError(e, sinks[i])
    mcastTable[i](x, sinks, handleError)
  }
}
function mcast2(x, sinks, handleError) {
  let i = 2
  try {
    sinks[--i].next(x)
    sinks[--i].next(x)
  } catch (e) {
    handleError(e, sinks[i])
    mcastTable[i](x, sinks, handleError)
  }
}
...
```

The problem is that the codebase size starts to grow rapidly. But as you can see, the code just repeats itself. And from the `combine` case we learned that evaluated functions are as fast as "hard coded" ones. Time to use `eval` again:
```js
const mcastTable = times(N, i => 
  eval(`(function mcast${i}(s,x,h,t) { var i=${i}; try{${times(i, i => "s[--i].next(x);").join("")}}catch(e){h(e,s[i]);i>0&&t[i](s,x,h,t);} })`));

function multicastN<T>(s: Array<Sink<T, any>>, x: T) {
  let n = s.length, i: number;
  for (i = 0; i < n; i++) safeNext(s[i], x);
} 
```
And now the `MulticastingSink` implementation can be changed to:
```typescript
class MulticastingSink<A, B> implements Sink<A, B>, ErrorHandler {
  private fn: (sinks: Array<Sink<A, any>>, x: A, handler: (err: any, s: Sink<A, B>) => void, table: Array<any>) => void; 
  constructor(private sinks: Array<Sink<A, B>>) {
    this.fn = this.sinks.length < 10 ? mcastTable[this.sinks.length] : multicastN;
  }
  next(x: A) {
    const mnext = this.fn 
    mnext(this.sinks, x, handleError, mcastTable);
  }
  // ...
}
```

As a result, we can implement the multicasting logic by using **only one** `try-catch` block, but still preserve the error handling.

And finally here are the performance results from the multicasting test.

Current implementation:
```
perf (master) $ node multicast.js 
multicast -> combine(addAll)100000 x 4 integers
-----------------------------------------------
xstream      14.16 op/s ±  2.00%   (65 samples)
most         55.90 op/s ±  1.12%   (66 samples)
rx 5          6.36 op/s ±  1.94%   (34 samples)
-----------------------------------------------
```
Optimized implementation:
```
perf (perf *) $ node multicast.js 
multicast -> combine(addAll)100000 x 4 integers
-----------------------------------------------
xstream      43.64 op/s ±  1.26%   (66 samples)
most         53.65 op/s ±  1.16%   (79 samples)
rx 5          7.44 op/s ±  2.14%   (39 samples)
-----------------------------------------------
```

And for comparison, here are results if one uses `.multicast` for each `most` operation:
```js
    var m1 = most.just(most.from(a).multicast()).delay(0).multicast().switch().multicast();
    var m2 = m1.map(x => x + 1).multicast();
    var m3 = m1.map(x => x + 2).multicast();
    var m4 = m1.map(x => x + 3).multicast();
    runners.runMost(deferred, 
      most.combineArray(addAll, [m1, m2, m3, m4]).multicast().drain());
```
Of course this is unfair comparison 'cause `most` is not designed to be used like that, but it makes the semantics of `xstream` and `most` somehow similar and it's interesting to see how well `most` performs in such case.
```
perf (perf *) $ node multicast.js 
multicast -> combine(addAll)100000 x 4 integers
-----------------------------------------------
xstream      43.44 op/s ±  1.38%   (66 samples)
most         35.28 op/s ±  1.23%   (78 samples)
rx 5          7.03 op/s ±  1.14%   (37 samples)
-----------------------------------------------
```

staltz:
@milankinen some good ideas there, specially I hadn't considered using eval to avoid the switch case.

Because some of these optimizations apply also to the old architecture, I tried the O(n) => O(1) fix on combine, and got some improvements already, check commit https://github.com/staltz/xstream/commit/ee4ec4c5133b7aa4360a22040966f4ab1fe7fb7b.

From
```
combine(add3) -> filter 500000 x 3 integers
-----------------------------------------------
xstream       8.56 op/s ± 13.42%   (24 samples)
most          3.94 op/s ±  2.62%   (23 samples)
rx 5          9.02 op/s ±  0.90%   (25 samples)
-----------------------------------------------
```

to
```
combine(add3) -> filter 500000 x 3 integers
-----------------------------------------------
xstream      12.31 op/s ±  1.11%   (32 samples)
most          3.90 op/s ±  1.28%   (23 samples)
rx 5          9.70 op/s ±  1.18%   (27 samples)
-----------------------------------------------
```

In my opinion, it would be easier to review and merge several PRs, one for each of these concerns, and we anyway needed special attention to combine, separately. This PR is a big project and will take some time to review, study, merge, make all tests pass (there are more tests in `master` branch since you started this), etc. But if you are willing to continue with the big project, go on. :)

milankinen:
👍 for the optimization.

That said, optimizing the initial check gives actually only little or almost no performance gain in real life cases, because the values usually arrive (almost) simultaneously to the `combine` function. The benchmark results are a bit biased because of the synchronous subscription and synchronous nature of `fromArray` producer.

If you want to gain easy performance improvements, I'd recommend the invoke function caching (and invoke function table). It gave me huge performance improvements (almost ten op/s) and applies to real-life cases as well. 

>In my opinion, it would be easier to review and merge several PRs, one for each of these concerns, and we anyway needed special attention to combine, separately.

I'd like to keep this pull request as a "meta PR" that gathers all of the information and notices in one place. And I'd also like to ensure that all features are possible to implement with this new call stack based approach before starting any merging. 

I agree that smaller PRs are easier to review and merge. My intention is to create smaller PRs based on this one if this PR reaches the point that something could actually be merged into the master.

staltz:
Perfect, let's do like that, and good point about the invoke table, we should do that.

monfera:
Epic work, guys! I really appreciate most.js and this PR is fantastic in that it reveals the process of how experts optimize things and how collaboration and transfer of ideas work. @briancavalier is there a writeup or 'making of' for most.js, from which the stack based approach is derived? [The architecture wiki](https://github.com/cujojs/most/wiki/Architecture) doesn't really go into that.

@milankinen @briancavalier A quick question on the call stack based event propagation that makes most.js (and this branch) fast: is it pretty much analogous to what is described in Rich Hickey's talk on transducers? I think part of the point with that was something analogous unless I'm missing something. Also, he mentioned something like, you can throw out around 90% of an FRP library and replace it with transducers - he must not have the most compact things like xstream, most and flyd in mind but it is good that flyd, kefir etc. provide transducer integration. I'd appreciate a one-sentence elucidation on commonality/difference on the stack based technique across most.js vs. transducers though might just look at it when I have the time. Btw. I made an experimental core of something similar to @milankinen's flyd, which, unlike his implementation, didn't reestablish the dependency graph each time, using instead a cached array first, then moving on to a stack based solution to avoid looping through the array of functions and got a nice perf increase. 

@juhovh I'm interested in most.js and xstream speedup for the same reasons I experimented with flyd. While the DOM - the most frequent target of cycle and thus xstream for now - is indeed slow, and so can be model calculations not to mention network operations, there *are* interesting use cases with WebGL, canvas, game development, or if someone goes against e.g. @staltz's advice to keep streams at a minimum and uses FRP inspired libraries in a very granular way, e.g. many hundreds of streams. So I wouldn't dismiss speedup even if the *typical* use case doesn't benefit too much from it.

paldepind:
> @briancavalier is there a writeup or 'making of' for most.js, from which the stack based approach is derived? The architecture wiki doesn't really go into that.

Such a thing would without a doubt be very interesting. @briancavalier did a truly impressive job on Most.js.

monfera:
@milankinen interesting observation with eval, V8 - I hope Google etc. keep it this way, despite the increasing use of transpilation, Babel/Bublé/Wasm/Elm etc.. Have you tried [Function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function) to see if it gives the same performance without `eval`?

milankinen:
Closing this PR due to lack of time.

PR #19, opened 2016-04-28
fix(core): fix filter fusion logic.

TylorS:
The logic to filter fusion was previously flawed.
.filter().filter() could never emit a value. as the second filter
would always be given a boolean, which is likely not the value it expects.
.map().filter() => FilterMap is also flawed logic as the filter is instead being
applied to the value *before* the map function which is likely not correct in most
circumstances.


staltz:
@TylorS do you have a test case for this?

TylorS:
@staltz I do now :)

staltz:
I had to cherry-pick this PR's commits to avoid a non-linear log graph, but it's merged now. Check 282d266f026cb33170b628785e8d16035f75f16d

Issue #11, opened 2016-04-26
Runtime error when using endWhen


AlexGalays:
When a stream using `endWhen` completes, an error is thrown inside this condition:

Stream.prototype._x
```
if (this._prod)
  this._prod._stop();
```

because it is assumed a producer always have a stop method which is not the case for `EndWhenOperator ` (and possibly a few other operators)

staltz:
Hmm, thanks for reporting.

staltz:
@AlexGalays can I kindly ask you to show some small reproducible case?

staltz:
Because EndWhenOperator actually does have a _stop() method: https://github.com/staltz/xstream/blob/670a0865838561e1a0d7cd4408b90ad22a35fb7d/src/core.ts#L414

AlexGalays:
You are right!

I actually had a custom producer missing a stop function in my own code; I somehow thought these were optional.

Cheers and sorry for the false alarm.

Widdershin:
Perhaps we should add errors if someone passes a producer missing a stop function?

staltz:
@Widdershin if it's done in `Stream.create()` and not in the constructor, then yes :+1: (because if it's in the constructor it would probably significantly hit performance)

Widdershin:
Sweet, I'll look into this 😄

AlexGalays:
To be honest, it was my fault for not using typescript :p the Producer interface shows both properties are mandatory.

Do you want to cater to JS programs with runtime checks?

Widdershin:
I feel the majority of users of xstream will be using JS. If you've run into this issue, others undoubtedly will. A little runtime cost can save a lot of time and pain for users and maintainers.

staltz:
Yeah, as long as it doesn't hit performance in a visible way, we should have them.

PR #22, opened 2016-04-29
feat(create): Throw an error if an producer is passed without start and stop

Widdershin:
As discussed in #11.

Is this what you were thinking @staltz?


TylorS:
LGTM :+1:

staltz:
Yep, looks all good to me. I'll merge it now, but it would be nice to do this for `createWithMemory` as well.

PR #42, opened 2016-06-02
feat(core): Only have 1 Type Stream [WIP]

TylorS:
This removes the need for a separate type MemoryStream.

This PR is a WIP, the goal is make a larger working distinction between streams with memory (Signals) and streams that do not have memory (Events).


TylorS:
This was an experiment, but as Staltz pointed out, not returning MemoryStream type when using TypeScript is a huge detriment to being able to use types to better understand your application and the flow of data through your application.

PR #43, opened 2016-06-03
feat(core): improve signature of operators regarding types

staltz:
This commit changes core operators to return either: (1) Stream only, (2) Stream or MemoryStream,
(3) MemoryStream only. Operators like fold() and startWith() now always return a MemoryStream, while
operators like map(), take(), and endWhen() return the same type that the source/instance stream
was. E.g. MemoryStream mapped will return a MemoryStream.

BREAKING CHANGE:
debug() now returns a MemoryStream if the input was also a MemoryStream.
endWhen() now returns a MemoryStream if the input was also a MemoryStream.
fold() now returns always a MemoryStream, not Stream.
imitate() only works on conventional Stream, will throw error on MemoryStream.
map() now returns a MemoryStream if the input was also a MemoryStream.
mapTo() now returns a MemoryStream if the input was also a MemoryStream.
replaceError() now returns a MemoryStream if the input was also a MemoryStream.
startWith() now returns always a MemoryStream, not Stream.
take() now returns a MemoryStream if the input was also a MemoryStream.

@TylorS


src/core.ts
```diff
@@ -1046,6 +1046,10 @@ export class Stream<T> implements InternalListener<T> {
     }
   }
 
+  private ctor(): typeof Stream {
```
TylorS: Very clever! :+1:

tests/operator/combine.ts
```diff
@@ -54,4 +56,20 @@ describe('Stream.prototype.combine', () => {
...
+    done();
+  });
+
+  it('should return a Stream when combining a MemoryStream with a MemoryStream', (done) => {
+    const input1 = xs.periodic(50).take(4).remember();
+    const input2 = xs.periodic(80).take(3).remember();
+    const stream: Stream<number> = input1.combine((x, y) => x + y, input2);
+    assert.strictEqual(stream instanceof Stream, true);
+    done();
+  });
```
TylorS: How does this behavior work? I don't see any adjustments to `combine()`, and I don't know how it would know how to do this

staltz: - Operators that always return `Stream` were unchanged because the operator function does explicitly "new **Stream**(producer)"
- Operators that return Stream or MemoryStream depending on the input use `new (this.ctor())(producer)` but the return type is Stream, and then MemoryStream overrides each operator to cast the return type to be MemoryStream 
- Operators that return MemoryStream always do `new MemoryStream(producer)`

TylorS: :+1:

TylorS:
LGTM

PR #45, opened 2016-06-04
feat(Observable): [WIP] add Observable implementation

TylorS:
Implements bi-directional conversion between Stream and Observable
and implements a .subscribe() method which returns a Subscription
type as defined by the current Observable specification

TODO
- [ ] Add tests for subscribe()
- [ ] Add tests for Stream.prototype\[symbol.observable\]()
- [ ] Add tests for Stream.from()
- [ ] Add Observable Tests
- [ ] Add Observable.fromStream() Tests
- [ ] Add Subscription Tests
- [ ] Add Documentation

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,9 @@
     "url": "https://github.com/staltz/xstream/issues"
   },
   "homepage": "https://github.com/staltz/xstream#readme",
-  "dependencies": {},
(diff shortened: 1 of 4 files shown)


TylorS:
I'm going to abandon this for now actually. The Observable Spec is changing practically daily now

PR #48, opened 2016-06-09
refactor(doctest): Use markdown-doctest regexRequire to load extras

Widdershin:
Saves anyone having to manually add any in future


staltz:
Thanks!

PR #50, opened 2016-06-10
Add browser console playground

morhetz:
I'm used to play with the library and prototype at the browser console on it's site (e.g. [Ramda](http://ramdajs.com/docs/), [Moment](http://momentjs.com/), [Immutable](https://facebook.github.io/immutable-js/)). So the workflow is simple: check docs, open browser console, try it.
[xstream docs](http://staltz.com/xstream/) currently lacks this feature
This PR exposes `xstream` and `xs` to the global window object. It also adds nice console banner (shamefully based on the immutable one)

preview

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -2,3 +2,21 @@
   <div class="wrapper">
   </div>
 </footer>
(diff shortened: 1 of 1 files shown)


staltz:
@morhetz Good idea! Thanks! I'll review and merge it soon, latest on Monday

PR #55, opened 2016-06-15
Prevent stack overflow while detecting cycles in _onlyReachesThis

Hypnosphi:
See testcase. After detaching `take(1)` the subscription graph is
```
outside ---> first ---> second ---> take(4)
 \                \    /
  \                <---
   ---> take(2)
```

Then, after detaching `take(2)`, `outside` calls `_OnlyReachesThis` which runs into infinite loop between `first` and `second`. To prevent this, whole path trace is provided as second argument to `_onlyReachesThis`, so that cycles can be detected even if they don’t contain `this`.

### DIFF
--- a/src/core.ts
+++ b/src/core.ts
@@ -1022,7 +1022,7 @@ export class Stream<T> implements InternalListener<T> {
   // force it to end its execution and dispose resources. This method
   // assumes as a precondition that this._ils has just one listener.
   _pruneCycles() {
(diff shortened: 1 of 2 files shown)


staltz:
Merged with https://github.com/staltz/xstream/commit/b971cbc11798e42d9e11c1fcd9fa0c6f17f06c36.

Thanks @Hypnosphi, perfect PR.

Issue #47, opened 2016-06-08
combine breaks when passed two streams that share a parent


Widdershin:
Here is a runnable reproducer:

```js
import assert from 'assert';
import xs from 'xstream';

function Counter (click$) {
  const count$ = click$
    .fold((total) => total + 1, 0)
    .remember();

  return {
    count$,

    countString$: count$.map(count => count.toString())
  };
}

const click$ = xs.create();

const counter = Counter(click$);

const combinedStreams = xs.combine(
  (...vals) => vals,
  counter.count$,
  counter.countString$
);

const expectedValues = [
  [0, '0'],
  [1, '1']
];

combinedStreams.addListener({
  next(vals) {
    console.log('expected:', expectedValues[0], 'actual:', vals);
    assert.deepEqual(expectedValues.shift(), vals);
  },
  error(err) {
    throw err;
  },
  complete() {}
});

click$.shamefullySendNext();
```

```bash
 $ ./node_modules/.bin/babel-node test.js
expected: [ 0, '0' ] actual: [ 0, '0' ]
expected: [ 1, '1' ] actual: [ 1, '0' ]

AssertionError: [ 1, '1' ] deepEqual [ 1, '0' ]
```

We expect to see that the `count$` and `countString$` are equal, but only the `count$` increments on `click$`.

If you pass `countString$` as the first stream to combine, the opposite happens.

This use case is important for @cycle/collection, for combining multiple sinks on each item.

I'm unsure if this is a bug, or if I'm just misunderstanding something fundamental.

ntilwalli:
I think it's just a "glitch" (http://staltz.com/rx-glitches-arent-actually-a-problem.html).  I posted a webpackbin (http://www.webpackbin.com/41pBk6ZN-) which does a "withLatestFrom" approach instead of `combine` and it works.

staltz:
Yeah, it looks like a glitch.

Widdershin:
Not sure that this is legitimately a glitch. If it was a glitch, I would expect to see this output:

```js
[ 0, '0' ]
[ 1, '0' ]
[ 1, '1' ]
[ 2, '1' ] // or whatever
```

The actual output is:

```js
[ 0, '0' ]
[ 1, '0' ]
[ 2, '0'' ]
```

The second stream never emits values (after the first). As such, I think this is related to #60

ntilwalli:
[...]

PR #60, opened 2016-06-20
[WIP] Cold stream

TylorS:
This is a work in progress PR which attempts to introduce the ability to convert a stream to be cold/unicast.


Widdershin:
Curious, is there a particular use case you have in mind?

TylorS:
Hey @Widdershin, yes I do. Where I work, we're hoping to create, an open-source, a Cycle + XStream material design component library. It of course, will be opionated towards the way we as a team have come to agree on designing and using components, but we still think it will be very useful for others :)

With that said, one of the patterns we have found very useful is always being able to assume your components receive `sources` and **everything** within your `sources` object is a stream. e.g.

```js

function main (sources) {
  ...
  // a control is a Component that makes up part of your form
  const form = Form({...sources, controls$: xs.of([
    {field: 'fullName', control: FullNameInput},
    {field: 'email': control: EmailInput}
  ])})
  ...
}
```

The problem that we have run into, is that using with `xs.of()` returning a *synchronous* hot/multicasted stream, only the first listener ever gets the values.

```js

const stream = xs.of(1, 2, 3)

stream.addListener(logListner)
/* internally
calls stream._add(logListener)
ref count goes from 0 -> 1
effectively calls FromArrayProducer._start(this)
FromArrayProducer._start effectively calls
  - logListener.next(1)
  - logListener.next(2)
  - logListener.next(3)
  - logListener.complete()

Then lastly removes logListener for the internal set of listeners and calls for the asynchronous stop of FromArrayProducer
*/

/* now we come here */
stream.addListener(otherListener)
/* calls stream._add(otherListener)
The above cancels the stop of the original FromArrayProducer
*/
```

The issue here is that FromArrayProducer has never been stopped, but will also never emit a value ever again, leaving `otherListener` to effectively be listening to `xs.never()`

I'm not sure if there really is a way to unshare a stream in xstream without large refactoring though.

tests/extra/cold.ts
```diff
@@ -0,0 +1,58 @@
...
+        stream.removeListener(listener);
+        done();
+      }
+    };
+
+    stream.addListener(listener);
+  });
+
+  it('should restart the producer for every new listener', (done) => {
+    const stream = xs.periodic(100).take(3).compose(cold);
```
staltz: Here's why this test fails.
xs.periodic(100) is hot
xs.periodic(100).take(3) is hot
xs.periodic(100).take(3).cold is "cold" but behaves like hot
xs.of(0, 1, 2) is hot
xs.of(0, 1, 2).compose(cold) is **truly cold**
xs.of(0, 1, 2).compose(cold).take(3) is hot

Which means if we want cold behavior to propagate, we need to implement the `ctor()` technique in all core operators.

staltz: And... that may hurt reasonaboutability of xstream in general, because it's easier if we can assume that "*all streams are hot*" than to assume that "*it may be cold or hot, we don't know*".

I'm not sure what to do in this situation, but I'd like to explore first how this could be solved elegantly (no Promise resolve for instance) with xstream as it is.

TylorS:
Yeah.. confusion will definitively not be goal in this. Just hope we can find a solution that works nicely.

Hypnosphi:
Wait, but `complete()` stops the producer synchronously. It's not `_removeListener()`.

Hypnosphi:
`_c` stops the streams producer [synchronously](https://github.com/staltz/xstream/blob/master/src/core.ts#L956), so this shouldn't be the source of described problem. See example: http://www.webpackbin.com/EkXEAMGBW

Widdershin:
Nice explanation @TylorS. I think this problem is what I ran into in #47, which was dismissed as a glitch.

Now, correct me if I'm wrong, but I thought part of the reason why `xstream` was created was to avoid needing to think about `hot`/`cold` in the context of a Cycle app. Did this turn out to be a leaky abstraction after all?

staltz:
@Widdershin Too early to say. Sometimes I run into some confusion similar to this, just to realize I had to remember() some state$ (whereas RxJS doesn't require you to shareReplay in those cases because of coldness), but then I *should* be calling remember() to model "values over time" like state. So I'd say before we jump to that conclusion, let's look case by case and how to elegantly solve with today's xstream.

TylorS:
I don't think I'm going to be able to do this without much much more thought on how to accomplish this.

PR #61, opened 2016-06-23
fix(dist): smaller kB footprint

staltz:
By renaming variables and smartly initializing class members in constructors, we can chip off about
half a kilobyte. lol.


staltz:
@TylorS can you review and let me know what you think of this? I didn't write it directly to master because I'm not sure if 0.5kB is worth it.

staltz:
PS: in the diff viewer, ignore the dist files of course, look at core.ts.

Widdershin:
This could make errors more cryptic. Errors like `this.property is undefined` are bad enough, but `q.p is undefined` is much harder to understand.

TylorS:
@staltz I'm torn on this. I don't think we have even remotely useful error messages now, but this definitely would make them worse. The code looks fine though, but I don't know whether its worth it.

staltz:
Yeah I'm also not sure if this PR is worth just 0.5kB.
I'll let it rot here for a few more days.

staltz:
This PR = 💩

PR #70, opened 2016-07-05
Enable `remember` usage with `xs.create()`

### LINKED ISSUE #69: Remembered stream does not add listener when source is created directly via 'xs.create'
I noticed this odd behavior (what looks like a bug).  When `remember` is called directly on a raw stream created via `xs.create` that has no producer (i.e. which will emit based on the `shamefully*` methods), it doesn't add listeners since the `_prod` property is `undefined`.  
```
import xs from 'xstream'

const test$ = xs.create()

// This line works
//const oddBehavior$ = test$  

// This line does not work. Why?
const oddBehavior$ = test$.remember()

oddBehavior$.addListener({
  next: n => bin.log(`Next: ${n}`),
  error: e => bin.log(`Error: ${e}`),
  complete: () => bin.log(`Complete`)
})

test$.shamefullySendNext(1)
```
You can see the issue here: http://www.webpackbin.com/4JFCBWrL-

Hypnosphi:
Fixes #69


tests/operator/remember.ts
```diff
@@ -131,4 +131,24 @@ describe('Stream.prototype.remember', () => {
     });
     done();
   });
+
+  it ('should work properly with `xs.create()`', (done) => {
+    const stream = xs.create().remember();
```
ntilwalli: This should be changed to create a reference to the `xs.create` stream and another references to the external `remember` stream.  The `shamefullySendNext` should be on the internal stream.  So it should read `const stream = xs.create(); const external = stream.remember();`

tests/operator/remember.ts
```diff
@@ -131,4 +131,24 @@ describe('Stream.prototype.remember', () => {
...
+    stream.shamefullySendNext('foo');
+
+    let expected = ['foo', 'bar'];
+    stream.addListener({
```
ntilwalli: The should be changed to `external.addListener({`

staltz:
Looks good, just need to fix the issues Nikhil mentioned.

Hypnosphi:
Thank you, fixed this. Had to add one more listener to start the operator chain up.

tests/operator/remember.ts
```diff
@@ -131,4 +131,27 @@ describe('Stream.prototype.remember', () => {
     });
     done();
   });
+
+  it ('should work properly with `xs.create()`', (done) => {
```
staltz: Can you split this into 2 tests? One test that has `shamefullySendNext('foo')` before `remembered.addListener` and another that has every shamefullySendNext after the addListener?

Hypnosphi:
Oops, seems I'm doing something wrong with git

Hypnosphi:
Splitting done, anyway

staltz:
Thanks! Merged and will release in the next patch version soon

ntilwalli:
@Hypnosphi @staltz Very cool. Thanks!

Issue #65, opened 2016-07-03
fromEvent on node.js events


jfedyczak:
I'm trying to migrate from most.js. Is it possible to use `fromEvent` on node.js events? I'm trying to pass `EventEmitter` as a source but with no luck...

Hypnosphi:
As stated in [docs](https://github.com/staltz/xstream/blob/master/EXTRA_DOCS.md#fromEvent), `fromEvent` is suited for DOM events. It uses `.addEventListener` internally. `fromNodeEvent` could be implemented like this:
```js
function fromNodeEvent(emitter, event) {
  let callback;
  return xs.create({
    start(listener) {
      callback = value => listener.next(value);
      emitter.addListener(event, callback);
    },
    stop() {
      emitter.removeListener(event, callback);
    }
  });
}
```

jfedyczak:
Are you planning on adding such method? It is present in `most.js` and in `RxJS`. I would provide a PR, but my TypeScript is not that fluent...

staltz:
I think we could do this, since fromEvent is an extra, and size is not a concern, but parity with RxJS and most.js is a benefit.

jfedyczak:
Thank you for the fix, but I'm trying to test this and I believe there's a problem with emitting events with more than one argument as in:

	var fromEvent = require('xstream/extra/fromEvent').default
	var http = require('http')

	var srv = http.createServer()

	fromEvent(srv, 'request')
	.addListener({
	  next: function(req, res) { // <-- problem here
		  // res is undefined
		  res.end(req.url)
	  },
	  error: function(err) { console.error(err) },
	  complete: function() { console.log('completed') },
	})

	srv.listen(3002)

Possible solution would be emitting an array with all the args or passing all the args...

staltz:
Yeah, it would have to be an array or an object.

xtianjohns:
I sympathize with the use case there, @jfedyczak. I think implementations in other libraries handle this issue with selector support, which allows users to declare how to map the arguments from the emitter into the stream.

I'd be happy to take a stab at that, but I think it deserves a separate issue and maybe a discussion about how that approach fits in with the larger goals of xstream.

PR #73, opened 2016-07-09
Support NodeJS Event Emitters in fromEvent extra

xtianjohns:
Add support for stream generation from EventEmitters. 

## Description
`fromEvent` will accept as a first argument either a DOMElement or an EventEmitter. When providing `fromEvent` with an EventEmitter, the third argument (useCapture) is always ignored.

There are two potentially outstanding concerns.

**First**: This is my first contribution to xtream and I'm not familiar with how the documentation is built. I've updated JSDOC comments with new examples, but it's possible that my comments don't work with the documentation generator.
**Second**: Parity with RxJS and Most.js should probably involve a change to the third parameter here. I believe they support an object with configuration options, rather than a boolean. I'm not certain if that work would be a requirement to merge, but a brief discussion probably couldn't hurt. I'm happy to update the implementation to accommodate any updated signatures.

## Motivation and Context
This should resolve staltz/xstream#65.

## How Has This Been Tested?
Tests have been split between the two possible uses of the extra and are passing.

## Checklist:
- [x] My code follows the code style of this project. (I think so?)
- [x] I have updated the documentation accordingly. (I've done some work, but need to confirm)

### DIFF
--- a/src/extra/fromEvent.ts
+++ b/src/extra/fromEvent.ts
@@ -1,4 +1,6 @@
-import {Stream, InternalProducer, InternalListener} from '../core';
+/// <reference path="../../typings/globals/node/index.d.ts" />
+import { EventEmitter } from 'events';
+import { Stream, InternalProducer, InternalListener } from '../core';
 
 export class DOMEventProducer implements InternalProducer<Event> {
   public type = 'fromEvent';
@@ -22,6 +24,25 @@ export class DOMEventProducer implements InternalProducer<Event> {
   }
 }
 
+export class NodeEventProducer implements InternalProducer<any> {
+  public type = 'fromEvent';
+  private listener: Function;
+
+  constructor(private node: EventEmitter, private eventName: string) { }
+
+  _start(out: InternalListener<any>) {
+    this.listener = (e: any) => out._n(e);
+    const {node, eventName} = this;
+    node.addListener(eventName, this.listener);
+  }
+
+  _stop() {
+    const {node, eventName, listener} = this;
+    node.removeListener(eventName, listener);
+    this.listener = null;
+  }
+}
+
 /**
  * Creates a stream based on DOM events of type `eventType` from the target
  * node.
@@ -29,11 +50,11 @@ export class DOMEventProducer implements InternalProducer<Event> {
  * Marble diagram:
  *
  * ```text
- *   fromEvent(node, eventType)
+ *   fromEvent( element, eventType )
  * ---ev--ev----ev---------------
  * ```
  *
- * Example:
+ * Examples:
  *
  * ```js
  * import fromEvent from 'xstream/extra/fromEvent'
@@ -45,7 +66,7 @@ export class DOMEventProducer implements InternalProducer<Event> {
  *   next: i => console.log(i),
  *   error: err => console.error(err),
  *   complete: () => console.log('completed')
- * })
+ * });
  * ```
  *
  * ```text
@@ -54,15 +75,37 @@ export class DOMEventProducer implements InternalProducer<Event> {
  * > 'Button clicked!'
  * ```
  *
- * @param {EventTarget} node The element we want to listen to.
- * @param {string} eventType The type of events we want to listen to.
- * @param {boolean} useCapture An optional boolean that indicates that events of
+ * ```js
+ * import fromEvent from 'xstream/extra/fromEvent'
+ * import {EventEmitter} from 'events';
+ *
+ * const MyEmitter = new EventEmitter();
+ * const stream = fromEvent( MyEmitter, 'foo' );
+ *
+ * stream.addListener({
+ *   next: i => console.log(i),
+ *   error: err => console.error(err),
+ *   complete: () => console.log('completed')
+ * });
+ *
+ * MyEmitter.emit( 'foo', 'bar' );
+ * ```
+ *
+ * ```text
+ * > 'bar'
+ * ```
+ *
+ * @param {EventTarget|EventEmitter} element The element upon which to listen.
+ * @param {string} eventName The name of the event for which to listen.
+ * @param {boolean?} useCapture An optional boolean that indicates that events of
  * this type will be dispatched to the registered listener before being
  * dispatched to any EventTarget beneath it in the DOM tree. Defaults to false.
  * @return {Stream}
  */
-export default function fromEvent(node: EventTarget,
-                                  eventType: string,
-                                  useCapture: boolean = false): Stream<Event> {
-  return new Stream<Event>(new DOMEventProducer(node, eventType, useCapture));
+export default function fromEvent( element: EventTarget | EventEmitter, eventName: string, useCapture: boolean = false): Stream<Event|any> {
+  if ( element instanceof EventEmitter ) {
+    return new Stream<any>(new NodeEventProducer( element, eventName ));
+  } else {
+    return new Stream<Event>(new DOMEventProducer( element, eventName, useCapture ));
+  }
 }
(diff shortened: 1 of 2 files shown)


staltz:
Really good PR. I'd fix just some minor issues, usually I do this myself after merging, but you seem willing to do that. I'll comment inline. 👍

src/extra/fromEvent.ts
```diff
@@ -1,4 +1,6 @@
-import {Stream, InternalProducer, InternalListener} from '../core';
+/// <reference path="../../typings/globals/node/index.d.ts" />
+import { EventEmitter } from 'events';
+import { Stream, InternalProducer, InternalListener } from '../core';
```
staltz: Code style: we use no spaces surrounding curly braces. So, `{EventEmitter}`.

src/extra/fromEvent.ts
```diff
@@ -22,18 +24,37 @@ export class DOMEventProducer implements InternalProducer<Event> {
...
 
+export class NodeEventProducer implements InternalProducer<any> {
+  public type = 'fromEvent';
+  private listener: Function;
+
+  constructor(private node: EventEmitter, private eventName: string) { }
+
+  _start(out: InternalListener<any>) {
+    this.listener = (e: any) => out._n(e);
+    const {node, eventName} = this;
```
staltz: Because xstream is written in an extreme manner to shrink code size as much as possible, it's actually not worth doing the `const {node, eventName} = this` trick in this case, because we use `node` just once below. So we'd use just `this.node` directly. However, if we used `node` twice or more, then it'd be worth doing the trick.

src/extra/fromEvent.ts
```diff
@@ -22,18 +24,37 @@ export class DOMEventProducer implements InternalProducer<Event> {
...
+    node.addListener(eventName, this.listener);
+  }
+
+  _stop() {
+    const {node, eventName, listener} = this;
```
staltz: Same here.

src/extra/fromEvent.ts
```diff
@@ -22,18 +24,37 @@ export class DOMEventProducer implements InternalProducer<Event> {
...
+
 /**
  * Creates a stream based on DOM events of type `eventType` from the target
  * node.
  *
  * Marble diagram:
  *
  * ```text
- *   fromEvent(node, eventType)
+ *   fromEvent( element, eventType )
```
staltz: No surrounding spaces

src/extra/fromEvent.ts
```diff
@@ -45,7 +66,7 @@ export class DOMEventProducer implements InternalProducer<Event> {
  *   next: i => console.log(i),
  *   error: err => console.error(err),
  *   complete: () => console.log('completed')
- * })
+ * });
```
staltz: No semicolon in examples

src/extra/fromEvent.ts
```diff
@@ -54,15 +75,39 @@ export class DOMEventProducer implements InternalProducer<Event> {
...
+ * const stream = fromEvent( MyEmitter, 'foo' );
+ *
+ * stream.addListener({
+ *   next: i => console.log(i),
+ *   error: err => console.error(err),
+ *   complete: () => console.log('completed')
+ * });
+ *
+ * MyEmitter.emit( 'foo', 'bar' );
+ * ```
```
staltz: Good example.
Could remove the semicolons though.

src/extra/fromEvent.ts
```diff
@@ -22,18 +24,37 @@ export class DOMEventProducer implements InternalProducer<Event> {
...
+  _stop() {
+    const {node, eventName, listener} = this;
+    node.removeListener(eventName, listener);
+    this.listener = null;
+  }
+}
+
 /**
  * Creates a stream based on DOM events of type `eventType` from the target
  * node.
```
staltz: Maybe add a mention of "either DOM events or EventEmitter events" here.

src/extra/fromEvent.ts
```diff
@@ -54,15 +75,39 @@ export class DOMEventProducer implements InternalProducer<Event> {
...
  * @return {Stream}
  */
-export default function fromEvent(node: EventTarget,
-                                  eventType: string,
-                                  useCapture: boolean = false): Stream<Event> {
-  return new Stream<Event>(new DOMEventProducer(node, eventType, useCapture));
+export default function fromEvent( element: EventTarget | EventEmitter,
+                                   eventName: string,
+                                   useCapture: boolean = false): Stream<Event|any> {
+  if ( element instanceof EventEmitter ) {
```
staltz: No space around the parenthesis.
Also, I'd rather use duck typing here instead of `instanceof`. I know it's probably ok for a vast majority of use cases (I can't come up with an example where we'd cross iframe boundaries), but our general approach in xstream is to use duck typing, so if we'd had just this one working differently, it'd be a surprise for someone at some point.

src/extra/fromEvent.ts
```diff
@@ -54,15 +75,39 @@ export class DOMEventProducer implements InternalProducer<Event> {
...
+    return new Stream<any>(new NodeEventProducer( element, eventName ));
+  } else {
+    return new Stream<Event>(new DOMEventProducer( element, eventName, useCapture ));
```
staltz: No space around parenthesis.
(yeah I know I need to make TSLint do this for us, sigh)

tests/extra/fromEvent.ts
```diff
@@ -103,3 +136,56 @@ describe('fromEvent (extra)', () => {
...
 });
+
+describe('fromEvent (extra) - EventEmitter', () => {
+  it('should call addListener with expected parameters', () => {
+    const target = new FakeEventEmitter();
+    const stream = fromEvent(target, 'test');
+
+    stream.addListener({next: noop, error: noop, complete: noop});
+
+    assert.strictEqual('test', target.event);
```
staltz: These are good tests. However over here there's some Yoda talk going on (e.g. "if monday is today", as opposed to "if today is monday"). The API is `assert.strictEqual(actual, expected)`, so it should be `assert.strictEqual(target.event, 'test')`

staltz:
PS: I'm fine with merging and doing the adjustments myself, so let me know what you prefer.

src/extra/fromEvent.ts
xtianjohns: Got it. Rx tests for `addListener` method. If that seems reasonable, I'll go that route.

xtianjohns:
Thank you, @staltz. I'll fix these up.

xtianjohns:
Apologies for the whitespace issues: I had my linter disabled. I've done my best to address each of your comments.

Regarding the type guards inside the extra, I used Rx as the model. The guards are slightly different between [4.x](https://github.com/Reactive-Extensions/RxJS/blob/master/src/core/linq/observable/fromevent.js#L86) and [5.x-beta](https://github.com/ReactiveX/rxjs/blob/master/src/observable/FromEventObservable.ts#L11-L13), so I implemented the one I considered simpler. I understand that the name, internal logic, and implementation of that guard might merit a revisit in the future. For now, though, it should be just as robust as `4.x`.

src/extra/fromEvent.ts
```diff
@@ -11,29 +13,49 @@ export class DOMEventProducer implements InternalProducer<Event> {
...
+  }
+
+  _stop() {
+    this.node.removeListener(this.eventName, this.listener);
+    this.listener = null;
+  }
+}
+
+function isEmitter(element: any): boolean {
+    return element.addListener;
```
staltz: `addListener` and also `emit`, just because xstream Streams also have `addListener`. :)

staltz:
Thanks a lot! Merged. Will release a new version soon.

PR #76, opened 2016-07-10
fix xs.merge completion

### LINKED ISSUE #74: xs.merge() bug - fixed by debug()
Here I pass in a stream of actions to my component but my component is not initialized correctly when I use xs.merge. If I don't it is initialized fine:
http://www.webpackbin.com/E1oZ4rc8Z

Here we see that adding -debug() fixes the issue:
http://www.webpackbin.com/VybPuSqLZ

Hypnosphi:
This is intended to fix #74 

The problem it that case is that `MergeProducer#_stop` method never gets called because of cancelled async stoppage. So when original stream gets executed second time and then completed, the `ac` counter goes to -1 and that doesn't lead to merged stream completion.

The idea of PR is that we can be sure about (virtual) count of active listeners right before we iterate through `insArr` and call `_add`s, because each of the incoming streams either runs already or gets started by `_add`.

There's still a potential problem in case if some incoming stream gets restarted by another listener. To fix that, `merge` should be rewritten in the manner of `combine` to track which of the streams actually completes.

### DIFF
--- a/src/core.ts
+++ b/src/core.ts
@@ -83,16 +83,16 @@ function and<T>(f1: (t: T) => boolean, f2: (t: T) => boolean): (t: T) => boolean
 export class MergeProducer<T> implements Aggregator<T, T>, InternalListener<T> {
   public type = 'merge';
   public out: Stream<T> = null;
-  private ac: number; // ac is activeCount, starts initialized
+  private ac: number; // ac is activeCount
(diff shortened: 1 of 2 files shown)


staltz:
Thanks a lot @Hypnosphi , this was perfect fix.

PR #79, opened 2016-07-13
feat(SubscribableProducer): replace Producers with Subscribables [WIP]

### LINKED ISSUE #7: fromObservable() / toObservable()
The ES Observable https://github.com/zenparsing/es-observable/, using perhaps this https://www.npmjs.com/package/symbol-observable

Then send PR to https://github.com/sindresorhus/awesome-observables

### LINKED ISSUE #72: Replace Producer concept with (ES or RxJS) Observable
Xstream is hot streams only, but in rare cases we need cold streams. @TylorS had experienced this hard limitation in a convincing use case. We were discussing what alternative solutions do we have, and we landed on one very interesting possible solution: *to replace Producers in xstream with typical Observables*.

That would mean dropping the `{start, stop}` API for a familiar `subscribe` + `subscription.unsubscribe` API. It means everywhere where we expect a Producer, we would expect a cold Observable, e.g. `xs.create(observable)`, where `observable` is an object with a `subscribe(observer)` function.

This would enable also very interesting interop with RxJS, because we could do `xs.create(Rx.Observable.of(1,2,3))` for instance. The only requirement that we have from `obs` in `xs.create(obs)` would be that it's an object with `subscribe(observer)` function. Alternatively, we could also support the API `xs.create(subscribe)` where `subscribe` is the function `subscribe(observer)`. So the type signature for `create` would be:

```js
type Subscribe<T> = (observer: Observer<T>) => Subscription;

interface Observable<T> {
  subscribe: Subscribe<T>;
}

function create<T>(o: Observable<T> | Subscribe): Stream<T>;
```

In essence, `xs.create` would behave like a multicast in RxJS, with the additional effect of converting from cold world to hot world **permanently**, since we don't have a conversion path from xs.Stream => Observable.

This is for now very experimental, @TylorS will give it a try. He's going to build it as an experiment in a PR. It would affect a lot of the current infrastructure in xstream, and would mean a breaking change, **but** for most xstream/Cycle.js users there wouldn't be any noticeable changes if they are using streams created in drivers and just using operators inside `main()`.

Anyone else wants to give their opinion about this? Any foreseeable problems that I haven't mentioned? @Widdershin @Hypnosphi @laszlokorte @blesh

TylorS:
# Still to be done

- [ ] Get tests passing, modifying tests **only** for the breaking change producer=>subscribable
- [ ] map+filter fusion still working
- [ ] Benchmark the perf, don't let it go too low
- [ ] Put all core stuff in `core.ts`
- [ ] Use InternalListener not Listener for most code in core.ts
- [ ] Merge FooOperator with FooListener where applicable (?)
- [ ] Make sure operators have `ins` and `out`, make sure aggregators have `insArr` and `out`
- [ ] Use `NO` object instead of `null`, like in existing code in master branch
- [ ] Manual code minification (naming of vars)

-----

This is absolutely a Work In Progress.

This attempts to provide greater interoperation between other existing Stream libraries which implement the Subscribable interface.

```typescript
interface Subscribable<T> {
  subscribe (listener: Listener<T>): Subscription
}

interface Listener<T> {
  next (x: T): void
  error (e: any): void
  complete (): void
}

interface Subscription {
  unsubscribe (): void
}
```
This allows for ability to use the operators and factories from other Stream
libraries while still getting the ease-of-use of XStream, and also provides the ability to use Cold/Unicast type Streams if required (though not often).

Examples
```js
import xs from 'xstream'
import { iterate } from 'most'

const stream = xs.create(iterate(x => x + 1, 0)).take(3)

stream.addListener({
  next: x => console.log(x),
  error: e => console.error(e),
  complete: () => console.log('complete')
}) // logs 0, 1, 2, 'complete'
```

```js
import xs from 'xstream'
import { Observable } from 'rxjs'

const stream = Observable.from( xs.of(1, 2, 3) )
```

closed #7 
closes #72 

BREAKING CHANGES:
  Before: `xs.create({ start(listener: Listener): void, stop (): void })`

  After: `xs.create({ subscribe(listener: Listener): Subscription })`


TylorS:
@staltz This is only broken out into the `src/operator` folder while I'm working on this, just for the ease of working on it. It'll get merged back into `core.ts` when I've got all the tests passing.

package.json
```diff
@@ -9,7 +9,7 @@
     "changelog": "conventional-changelog --infile CHANGELOG.md --same-file --release-count 0 --preset angular",
     "lint": "tslint -c tslint.json src/**/*.ts",
     "premocha": "npm run compile",
-    "mocha": "mocha tests/*.ts tests/**/*.ts --require ts-node/register",
+    "mocha": "mocha tests/*.ts tests/factory/*.ts tests/operator/*.ts --require ts-node/register",
```
staltz: What were you solving with this change?

src/core.ts
```diff
@@ -1,133 +1,90 @@
...
+class MergeProducer<T> implements Subscribable<T> {
+  constructor (private streamsArray: Stream<T>[]) {}
 
-function compose2<T, U>(f1: (t: T) => any, f2: (t: T) => any): (t: T) => any {
-  return function composedFn(arg: T): any {
-    return f1(f2(arg));
-  };
-}
+  subscribe (listener: Listener<T>) {
+    const sa = this.streamsArray
```
staltz: Just to keep in mind: it's important that merge and combine implement the Aggregator interface because the field `insArr` will be used by devtools to traverse the stream graph.

staltz: And in normal operators, they should all have `ins` field for the same reason, for devtools.

src/core.ts
```diff
@@ -163,1775 +120,485 @@ export interface CombineSignature {
...
+  static throw (error: any) {
+    return new Stream<any>({
+      subscribe (listener: Listener<any>) {
+        listener.error(error)
+        return new StreamSubscription(noop)
+      }
+    })
+  }
+
+  static from<T> (x: ArrayLike<T> | Subscribable<T>) {
```
staltz: 👍 this is new

src/core.ts
```diff
@@ -163,1775 +120,485 @@ export interface CombineSignature {
...
 
-  _start(out: Stream<T>): void {
-    this.out = out;
-    this.o._add(this.oil = new OtherIL(out, this));
-    this.ins._add(this);
+  static merge<T> (...streams: Stream<T>[]): Stream<T> {
+    const l = streams.length
+    return l === 0 ? Stream.empty()
+      : l === 1 ? streams[0]
+      : new Stream<T>(new MergeProducer<T>(streams))
```
staltz: Smart optimization. 👍

package.json
TylorS: I haven't written the extra operators, so compiling the tests fail, it can be reverted when the extra operators are present again.

src/core.ts
```diff
@@ -163,1775 +120,485 @@ export interface CombineSignature {
...
-    return <MemoryStream<T>> super.endWhen(other);
-  }
-
-  replaceError(replace: (err: any) => Stream<T>): MemoryStream<T> {
-    return <MemoryStream<T>> super.replaceError(replace);
-  }
-
-  debug(labelOrSpy?: string | ((t: T) => void)): MemoryStream<T> {
-    return <MemoryStream<T>> super.debug(labelOrSpy);
+export class CompositeSubscription implements Subscription {
```
staltz: In RxJS v5, their `Subscription` is already a composite subscription. I'd recommend we call our CompositeSubscription just `Sub` (saving space on character count is a thing we do *shrug*).

src/operator/drop.ts
```diff
@@ -0,0 +1,20 @@
+import { Subscribable, Listener, Subscription } from '../interfaces'
+
+export class DropOperator<T> implements Subscribable<T> {
+  constructor (private amount: number, public _subscribable: Subscribable<T>) {}
+
+  subscribe (listener: Listener<T>): Subscription {
+    return this._subscribable.subscribe(new DropListener<T>(this.amount, listener))
+  }
```
staltz: Do you think it's possible to merge DropOperator and DropListener? Having them separate is useful in RxJS v5 where we have unicast by default, but considering that after we "multicast" using `xs.create(subscribable)`, from that point onwards we can assume hot only. I can't remember but I suspect that having these two merged in one is better for perf.

src/operator/drop.ts
```diff
@@ -0,0 +1,20 @@
...
+}
+
+class DropListener<T> implements Listener<T> {
+  constructor (private amount: number, private listener: Listener<T>) {}
+
+  next (x: T) {
```
staltz: Eventually we should implement InternalListener, then `next` -> `_n`.

src/operator/filter.ts
```diff
@@ -0,0 +1,24 @@
...
+export type Predicate<T> = (x: T) => boolean
+
+export class FilterOperator<T> implements Subscribable<T> {
+  constructor (private predicate: Predicate<T>,
+               public _subscribable: Subscribable<T>) {}
+
+  subscribe (listener: Listener<T>): Subscription {
+    return this._subscribable.subscribe(new FilterListener(this.predicate, listener))
+  }
+}
```
staltz: What I said for DropOperator/DropListener applies to most of these.

staltz:
Good.

I think there's no point in me reviewing everything on the nitpicking level, since it's WIP. Let's try to do the approach: (1) make it work, (2) make it right, (3) make it fast.

So something along these lines:
- [ ] Get tests passing, modifying tests **only** for the breaking change producer=>subscribable
- [ ] map+filter fusion still working
- [ ] Benchmark the perf, don't let it go too low
- [ ] Put all core stuff in `core.ts`
- [ ] Use InternalListener not Listener for most code in core.ts
- [ ] Merge FooOperator with FooListener where applicable
- [ ] Make sure operators have `ins` and `out`, make sure aggregators have `insArr` and `out`
- [ ] Use `NO` object instead of `null`, like in existing code in master branch
- [ ] Manual code minification (naming of vars)

I can help with many of these tasks, particularly the bottom ones, like using InternalListener and onwards.

src/operator/drop.ts
TylorS: I would actually say that merging the 2 classes would have a negative performance impact. I opted to follow the most.js architecture here. The reason I would suspect negative perf is because at the point you merge the 2 classes you have to capture the `listener` from subscribe as `this.listener`

```typescript
export class DropOperator<T> implement Subscribable<T>, Listener<T> {
  private class listener: Listener<T> = null
  ...
  subscribe(listener: Listener<T>) {
    this.listener = listener // the type of this.listener has changed a will cause de-opt
  }
}
```

staltz: Hmm interesting. Ok, let's see.
By the way, this listener parameter is `out`.

package.json
staltz: Oh, sure, correct.

xtianjohns:
Just a fly on the wall here, but curious about this work in relation to #7. Obviously, that's focused on symbol usage, but [this comment](https://github.com/staltz/xstream/issues/7#issuecomment-214431060) in particular seems relevant to moving to/from other Observable interfaces, as in @TylorS's examples in the OP.

TylorS:
@xtianjohns This will also close #7

Issue #80, opened 2016-07-15
Improve Stream.merge types


tp:
When using `Stream.merge` I found that I had to manually define the types since TS could not infer them with the current signature, which makes it of course error prone:

```ts
const merged = xs.Stream.merge<A | B>(a, b);
```

From what I have seen so far to get variadic-ish generically typed arguments supported, we could use a function signature like this:
```ts
static merge<T1, T2>(s1: Stream<T1>, s2: Stream<T2>): Stream<T1 | T2>;
static merge<T1, T2, T3>(s1: Stream<T1>, s2: Stream<T2>, s3: Stream<T3>): Stream<T1 | T2 | T3>;
[...]
```

(I am pretty sure this works in .d.ts files, not sure how to annotate a function written directly in TS code with this though)

Code like this is often generated for up to 20 or 30 arguments, and then provides full type safety.

Hypnosphi:
what is `Stream.merge`? Since 4.0.0, there's only `xs.merge`.

Anyway, why do you need to merge streams of different types?

staltz:
I agree with this suggestion, seems easy to do and we should do it.

xtianjohns:
I can jump on this if no one objects. Two questions:

1. How far should we go when overloading the method? Rx goes to [six](https://github.com/ReactiveX/rxjs/blob/master/src/operator/merge.ts#L77-L92), but I feel like I've seen them do more elsewhere.

2. Should we one-line those overloads or try to keep them to the width throughout the rest of the project?

staltz:
@xtianjohns good! We should do like Combine has: https://github.com/staltz/xstream/blob/master/src/core.ts#L135

PR #82, opened 2016-07-18
Introduce xs.merge type declaration

xtianjohns:
Support type declarations when calling `xs.merge()`. This change allows consumers to leverage TypeScript to verify type safety when merging streams. It also brings `merge` in line with other static methods which already support this feature.

## Description
Modify static `merge` method to implement a new interface: MergeSignature. This interface is constructed similarly to the CombineSignature interface. Export the new interface.

There are two areas I think merit discussion.

**First**: I'm not exactly sure what the intended behavior of this method is when called without arguments. The interface says a stream is returned that emits items of any type. And I suppose that's true. However, when I call this without any arguments, I get a stream that doesn't emit anything ever (or complete). Obviously my fault as the caller, but perhaps we want to do something else when it's called without arguments.

**Second**: I don't know how to test this or if it deserves a test. I imported core into a test file and was able to get type support for the method.

## Motivation and Context
This should resolve staltz/xstream#80.

## Checklist:
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.

### DIFF
--- a/src/core.ts
+++ b/src/core.ts
@@ -132,6 +132,37 @@ export class MergeProducer<T> implements Aggregator<T, T>, InternalListener<T> {
   }
 }
 
+export interface MergeSignature {
+  (): Stream<any>;
+  <T1>(s1: Stream<T1>): Stream<T1>;
+  <T1, T2>(
+    s1: Stream<T1>,
+    s2: Stream<T2>): Stream<T1 | T2>;
+  <T1, T2, T3>(
+    s1: Stream<T1>,
+    s2: Stream<T2>,
+    s3: Stream<T3>): Stream<T1 | T2 | T3>;
+  <T1, T2, T3, T4>(
+    s1: Stream<T1>,
+    s2: Stream<T2>,
+    s3: Stream<T3>,
+    s4: Stream<T4>): Stream<T1 | T2 | T3 | T4>;
+  <T1, T2, T3, T4, T5>(
+    s1: Stream<T1>,
+    s2: Stream<T2>,
+    s3: Stream<T3>,
+    s4: Stream<T4>,
+    s5: Stream<T5>): Stream<T1 | T2 | T3 | T4 | T5>;
+  <T1, T2, T3, T4, T5, T6>(
+    s1: Stream<T1>,
+    s2: Stream<T2>,
+    s3: Stream<T3>,
+    s4: Stream<T4>,
+    s5: Stream<T5>,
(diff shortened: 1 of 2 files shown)


src/core.ts
```diff
@@ -132,6 +132,37 @@ export class MergeProducer<T> implements Aggregator<T, T>, InternalListener<T> {
...
+    s4: Stream<T4>,
+    s5: Stream<T5>): Stream<T1 | T2 | T3 | T4 | T5>;
+  <T1, T2, T3, T4, T5, T6>(
+    s1: Stream<T1>,
+    s2: Stream<T2>,
+    s3: Stream<T3>,
+    s4: Stream<T4>,
+    s5: Stream<T5>,
+    s6: Stream<T6>): Stream<T1 | T2 | T3 | T4 | T5 | T6>;
+  (...stream: Array<Stream<any>>): Stream<any>;
```
staltz: I think the catch-all case should be
```typescript
<T>(...stream: Array<Stream<T>>): Stream<T>;
```
Because that's what we had previously in the Stream class, under `static merge` etc.

staltz:
> First: I'm not exactly sure what the intended behavior of this method is when called without arguments. The interface says a stream is returned that emits items of any type. And I suppose that's true. However, when I call this without any arguments, I get a stream that doesn't emit anything ever (or complete). Obviously my fault as the caller, but perhaps we want to do something else when it's called without arguments.

As far as this PR is concerned, I don't think you need to change anything. It would be a breaking change, and we don't know if we want to do that.

> Second: I don't know how to test this or if it deserves a test. I imported core into a test file and was able to get type support for the method.

Check this test for an inspiration: https://github.com/staltz/xstream/blob/master/tests/factory/combine.ts#L26

The trick is to define the type of the variable, e.g. `const combined: Stream<[string, string]>` at the same time as assigning to it. TypeScript compilation will fail (hence the test will fail) if the right-hand side of the assignment doesn't match the expected type on the left-hand side.

Nitpick: the commit message should follow our convention. Just use `npm run commit` instead of `git commit`. But anyway I can easily rename the commit myself.

PS: good PR. :)

xtianjohns:
No problem - I'll get those fixed up this afternoon. Thanks!

xtianjohns:
Alright, the catch all signature should look better now. I'm not sure what happened there.

Commit message should be improved, and when I amended the commit I threw in a test.

staltz:
Perfect! Thanks!

PR #83, opened 2016-07-18
Make `MemoryStream#remember` a true bypass

Hypnosphi:
See discussion in gitter: https://gitter.im/staltz/xstream?at=578bb1a7914c51592b338e54


staltz:
Makes much sense. Thanks

PS: for the next PR, try to make a git commit message that follows the convention. Just use `npm run commit` instead of `git commit`.

Issue #85, opened 2016-07-19
Filters are applied backward


atomrc:
When chaining `filter`s on a stream, they are executed backward at run time. 

eg
```javascript
const filtered$ = value$ 
    .filter(val => val.foo)
    //this filter is played before the one above ... causing an error
    .filter(val => val.foo.bar === "ok")
```

and here is a jsbin to reproduce the error: https://jsbin.com/noribadoxa/edit?html,js,console

staltz:
Thanks for reporting.

staltz:
Fixed in v5.2.4.

atomrc:
Awesome thanks :)

PR #86, opened 2016-07-19
fix(operators): fix filters execution order

atomrc:
This should fix https://github.com/staltz/xstream/issues/85


atomrc:
OK sorry guys, don't mind this PR it's an error :)

Issue #84, opened 2016-07-18
Events with more than one argument


jfedyczak:
As per suggestion of @xtianjohns I'm creating new issue conected with [this issue](https://github.com/staltz/xstream/issues/65). There's a problem with emitting events with more than one argument as in:

	var fromEvent = require('xstream/extra/fromEvent').default
	var http = require('http')

	var srv = http.createServer()

	fromEvent(srv, 'request')
	.addListener({
	  next: function(req, res) { // <-- problem here
		  // res is undefined
		  res.end(req.url)
	  },
	  error: function(err) { console.error(err) },
	  complete: function() { console.log('completed') },
	})

	srv.listen(3002)

Possible solution would be emitting an array with all the args or passing all the args...

TylorS:
@jfedyczak I don't think Server's are really a use case for xstream, as it's designed for (and only for) Cycle.js

jfedyczak:
@TylorS Well, it is just an example of use with EventEmmiter event which is now supported by xstream as an extra...

staltz:
xstream listener functions, just like ES Observable or RxJS, will *always* take just one argument. If there is an API that deviates from that, then we need to convert multiple args to an array or an object that collects them. I'd be happy to get a PR for this for `fromEvent`.

sculeb:
It would help if the fromEvent had selector support like rxjs. Then one could map multiple arguments of the event to an object or just use the argument he needs or w/e.

usecase:

`// mqtt-connection`
`client.subscribe(topic);`
`fromEvent(client, 'message', (topic, message) => message.toString())`
`.addListener({`
` next: val => console.log(val),`
` error: err => console.log(err),`
` completed: () => console.log('done')`
`});`

xtianjohns:
I'm excited to take a stab at supporting these. @jfedyczak, I worked on the `fromEvent` support for Event Emitters, and you're right that the current implementation lacks the same features as some other stream libraries. One concern was that supporting selectors using the same signature as some other libraries would be a breaking change. Specifically, the third argument (used for DOM event support) would need to be moved.

I can get a PR in for this, but if folks have concerns about a potentially breaking change for the extra and want that to be tackled in a particular way, let's discuss. My instincts are to move the (currently) third argument into an optional fourth argument, which would be a hash of options.

Does this seem reasonable?

staltz:
[...]

PR #89, opened 2016-07-22
feat(fromEvent): Aggregate multiple arguments

xtianjohns:
If provided with an EventEmitter, `fromEvent` streams will emit an array of values when the EventEmitter provides listeners with multiple arguments.

## Description
The previous implementation naively provided only the first argument, which made `fromEvent` incompatible with event interfaces that rely on providing multiple arguments to their listeners, such as Node's `http.Server`.

An alternative solution involved providing selector support. This wasn't implemented for several reasons.

  1. Modifying the function signature for the extra would be a breaking change.
  2. An expanded function signature would become 'crowded', and one of the architectural goals of xstream is to keep interfaces as small as they need to be to achieve their goals.
  3. Aggregating emitted events into an array is consistent with the implementation in Most.js, and achieves similar flexibility with Rx when providing a selector function to `.map()` on a result stream.

`fromEvent` will now emit mixed-types. **If consumers are not responsible for calling `.emit()` on the source emitter, they should implement appropriate guards to ensure they are dealing with an intended type.**

I think that it would be great if `fromEvent` could support typed EventEmitters so that downstream consumers could rely on TypeScript to warn them about guards, but EventEmitter isn't a generic type and that would take some rejiggering that I didn't think was warranted to implement this feature.

## Motivation and Context
This should resolve staltz/xstream#84.

## Checklist:

### DIFF
--- a/src/extra/fromEvent.ts
+++ b/src/extra/fromEvent.ts
@@ -29,7 +29,9 @@ export class NodeEventProducer implements InternalProducer<any> {
   constructor(private node: EventEmitter, private eventName: string) { }
 
   _start(out: InternalListener<any>) {
-    this.listener = (e: any) => out._n(e);
+    this.listener = (...args: Array<any>) => {
+        return (args.length > 1) ? out._n(args) : out._n(args[0]);
+    };
     this.node.addListener(this.eventName, this.listener);
   }
 
@@ -47,6 +49,10 @@ function isEmitter(element: any): boolean {
  * Creates a stream based on either:
  * - DOM events with the name `eventName` from a provided target node
  * - Events with the name `eventName` from a provided NodeJS EventEmitter
+ *
+ * When creating a stream from EventEmitters, if the source event has more than
+ * one argument all the arguments will be aggregated into an array in the
+ * result stream.
  *
  * Marble diagram:
  *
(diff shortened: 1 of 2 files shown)


src/extra/fromEvent.ts
```diff
@@ -29,7 +29,9 @@ export class NodeEventProducer implements InternalProducer<any> {
   constructor(private node: EventEmitter, private eventName: string) { }
 
   _start(out: InternalListener<any>) {
-    this.listener = (e: any) => out._n(e);
+    this.listener = (...args: Array<any>) => {
+        return (args.length > 1) ? out._n([...args]) : out._n(args[0]);
```
xtianjohns: I probably don't need to expand `[...args]` here. @staltz, if this bothers you I'll fix it.

staltz: Yeah, `[...array]` is the same as `array`. It's like `x + 1 - 1`. :)

src/extra/fromEvent.ts
```diff
@@ -48,6 +50,10 @@ function isEmitter(element: any): boolean {
  * - DOM events with the name `eventName` from a provided target node
  * - Events with the name `eventName` from a provided NodeJS EventEmitter
  *
+ * When creating a steam from EventEmitters, if the source event has more than
+ * one argument all the arguments will be aggregated into an array in the
+ * result stream.
+ *
```
staltz: 👍

src/extra/fromEvent.ts
```diff
@@ -96,6 +102,26 @@ function isEmitter(element: any): boolean {
...
+ *   error: err => console.error(err),
+ *   complete: () => console.log('completed')
+ * })
+ *
+ * MyEmitter.emit('foo', 'bar', 'baz', 'buzz')
+ * ```
+ *
+ * ```text
+ * > ['bar', 'baz', 'buzz']
+ * ```
```
staltz: 👍

tests/extra/fromEvent.ts
```diff
@@ -187,4 +187,25 @@ describe('fromEvent (extra) - EventEmitter', () => {
...
+      error: (err: any) => done(err),
+      complete: () => {
+        assert.strictEqual(expected.length, 0);
+        done();
+      }
+    });
+
+    target.emit( 'test', 1, 'foo', true );
+    target.emit( 'test', 2, 'bar', false );
+  });
```
staltz: 👍

src/extra/fromEvent.ts
xtianjohns: This will also be good information to have in the future, do you prefer amended/rebased stuff for changes like this, or smaller fix commits?

staltz: Amended is better

staltz:
Thanks! Merged

staltz:
Will be released in v5.3.0

PR #101, opened 2016-08-14
fix(dropRepeats): handle circular dependencies

chromakode:
Ran into a tricky behavior tonight when feeding @cycle/history into itself: if we have a circular dependency, and the event emitted by `dropRepeats` triggers another input event in the same stream, we can get into an infinite loop. The `dropRepeats` operator will recurse because it emits before it updates its sentinel value until it overflows the stack.

This test seems to model the case I hit accurately, but the way it checks for completion feels a bit inelegant -- any suggestions? I tried using `endWith(input.last())` but that appeared to cause `stream` to complete without emitting any events.

### DIFF
--- a/src/extra/dropRepeats.ts
+++ b/src/extra/dropRepeats.ts
@@ -29,10 +29,9 @@ export class DropRepeatsOperator<T> implements Operator<T, T> {
     const u = this.out;
     if (!u) return;
     const v = this.v;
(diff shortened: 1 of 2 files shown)


staltz:
Thanks @chromakode !

staltz:
Released 5.3.3 with this.

Issue #91, opened 2016-07-25
debug, map, flatten bug


staltz:
http://jsbin.com/hihuvuqupo/edit?html,js,console

ShayMatasaro:
@staltz 
this jsbin as well as the ones listed on the main cycle site in the documentation section all fail with : 
ReferenceError: xstream is not defined

mz3:
I have a stream that does not emit anything unless there is a .debug(noop) tacked on the end. Do you think it is related, or should I post another issue with detail?

staltz:
Just open another issue. We can't be sure yet it is the same bug.

staltz:
This should be fixed now in v6.0

PR #106, opened 2016-08-16
fix(core): teardown and stop producer before complete/error

staltz:
To fix bugs such as #91 and others where adding .debug() would magically solve the bug, we need to change the semantics around producer stop and stream completion/error. Previously, a stream would complete first, and stop the producer after. Now, the stream stops its producer just before completing. This was important to make the test 'Stream.prototype.flatten with map should not run multiple executions of the inner' pass.

BREAKING CHANGE: in this version, when a stream completes or errors, its producer has already been stopped. In previous versions, the stream first completes, propagates the complete to other
listeners and operators, and then its producer is stopped. You may barely notice this breaking change when updating your code. Most existing code will still work like before.

- - -

For review since it's a breaking change. Related to bug #91. The test below currently does not pass in v5.x.x:

```js
    it('should not run multiple executions of the inner', (done) => {
      const inner = xs.periodic(350).take(2).map(i => i * 100);
      const outer = xs.periodic(200).take(4);
      const stream = outer.map(() => inner).flatten();

      const expected = [0, 100];
      stream.addListener({
        next: (x: number) => {
          assert.equal(x, expected.shift());
        },
        error: (err: any) => done(err),
        complete: () => {
          assert.equal(expected.length, 0);
          done();
        }
      });
    });
```

The listener of `stream` sees the values `0`, `0`, `100`, `100`, `100`, `100`. There are multiple concurrent executions of `stream`.

cc @TylorS @Hypnosphi @ntilwalli

### DIFF
--- a/src/core.ts
+++ b/src/core.ts
@@ -1089,21 +1089,21 @@ export class Stream<T> implements InternalListener<T> {
     this._err = err;
     const a = this._ils;
     const L = a.length;
+    this._x();
     if (L == 1) a[0]._e(err); else {
       const b = copy(a);
       for (let i = 0; i < L; i++) b[i]._e(err);
     }
-    this._x();
   }
 
   _c(): void {
     const a = this._ils;
     const L = a.length;
+    this._x();
     if (L == 1) a[0]._c(); else {
       const b = copy(a);
       for (let i = 0; i < L; i++) b[i]._c();
     }
-    this._x();
   }
 
   _x(): void { // tear down logic, after error or complete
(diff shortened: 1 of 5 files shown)


ntilwalli:
I consider the new behavior to be intuitive. I'm actually not clear on why the current behavior doesn't trigger the same output on the given example.  In the example, the outer stream always switches away from the inner stream before it has had a chance to emit until the very last emission, when the inner stream is allowed to emit to completion.  Why does the order of signaling to upstream sources and downstream consumers matter here?

staltz:
It's hard to explain, but it has to do with the `_stop` of one operator removing itself as a listener to the stream above, and then as the last listener leaving, it schedules an async stop for the stream above. So because the outer stream emits quickly in flatten, it gets a few `_stop` scheduled, but also the restarting behavior kicks in which the producer's start is called even if the producer has already been started. This PR makes it harder (or impossible?) for a producer to start if it has already started, because we don't want parallel executions of the same stream (meaning parallel _start). I wish I could explain it better, but there's many tricky parts that triggered this bug.

staltz:
Merged. Thanks for reviewing @ntilwalli

PR #112, opened 2016-08-28
support type-guards in Stream's filter method

maiermic:
If a type-guard is passed to filter, you know that the returned stream contains
only elements of the type that is checked by the type-guard.


staltz:
Thanks @maiermic !

PR #118, opened 2016-09-15
feat(Observable): Adds Basic Observable interop

TylorS:
Implements basic interop that is compatible with most.js and RxJS 5. 
Add `xs.fromObservable(observable)` method that will take an Observable and return an xstream `Stream`.
Add `xs.from(input)` for Observable interop, which will convert Arrays, Observable, and Promises using their respective fromX method.
Add `subscribe` method to Stream that return a Subscription with `unsubscribe`
Add `Symbol.observable` ponyfill `symbol-observable` as a dependency to use the same as most.js and RxJS 5.
Add tests for all of the above.

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,6 +59,7 @@
     "markdox": "^0.1.10",
     "mkdirp": "^0.5.1",
     "mocha": "^2.4.5",
+    "most": "^1.0.3",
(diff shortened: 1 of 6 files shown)


staltz:
Merged! Thanks!

Issue #121, opened 2016-09-20
Errors should be thrown if no listeners are attached


kay-is:
[Steam.prototype._e](https://github.com/staltz/xstream/blob/master/src/core.ts#L1181) swallows errors if no Listener is attached.

It should probably just re-throw the received error, if it can't pass it on.

PR #122, opened 2016-09-20
Throw error if no listener is attached #121

kay-is:
This prevents `Stream` from swallowing errors if no `Listener` is attached.

### DIFF
--- a/src/core.ts
+++ b/src/core.ts
@@ -1189,6 +1189,7 @@ export class Stream<T> implements InternalListener<T> {
       const b = copy(a);
       for (let i = 0; i < L; i++) b[i]._e(err);
     }
+    if (!this._d && L == 0) throw this._err;
   }
 
   _c(): void {
--- a/tests/operator/imitate.ts
+++ b/tests/operator/imitate.ts
@@ -220,7 +220,7 @@ describe('Stream.prototype.imitate', () => {
       return xs.of(1).compose(delay<number>(20));
     }).flatten();
     proxyAction$.imitate(action$);
-    const expected = [0, 1, 2];
+    const expected = [0, 1, 2, 3];
 
     let errors: Array<any> = [];
     state$.addListener({


staltz:
Sorry, can you please check the tests? https://travis-ci.org/staltz/xstream/builds/161442594

You can run this also locally. The tests are consistently failing.

kay-is:
Ah, I thought maybe the test were wrong, haha.

    Stream.prototype.imitate should not propagate errors in a cycle

But well, makes sense, I want to propagate errors with this PR and the test that prohibits this fails.

@staltz what do you propose? Deleting the test, since it checks something I explicitly want (not swallowing errors)?

staltz:
@kay-is we should definitely not delete that test. We don't want errors to be propagated (infinitely) in a loop.

xtianjohns:
I actually think that these failures are the result of a bad test.

```ts
it('should not propagate errors in a cycle', (done) => {
    const proxyAction$ = xs.create<number>();
    const state$ = proxyAction$.fold((state, action) => state + action, 0);
    const action$ = state$.map(state => {
      if (state === 3) {
        throw new Error(':(');
      }
      return xs.of(1).compose(delay(20));
    }).flatten();
    proxyAction$.imitate(action$);
    const expected = [0, 1, 2];

    let errors: Array<any> = [];
    state$.addListener({
      next: (x: number) => {
        assert.equal(x, expected.shift()); // THIS IS A BUG
      },
      error: (err: any) => {
        errors.push(err);
      },
      complete: () => {
        done('complete should not be called');
      },
    });

    setTimeout(() => {
      assert.equal(errors.length, 1);
      assert.equal(expected.length, 0);
      done();
    }, 150);
  });
```
I've flagged the line above, and it represents (at least in part, I believe) one of the problems that's trying to get fixed in this PR. Specifically, **if you log inside this next method on `master`, the assertion _fails_.**

And because that exception is swallowed and never thrown, Mocha is none the wiser. So, @kay-is's fix looks like it is doing its job very well. One more change should be made to this test, and that is appending a `3` to the end of `expected`. When I make that change, my tests pass.

staltz:
Yes, you guys are absolutely correct. Thanks!

Issue #97, opened 2016-08-02
FAQ translation of RxJS' withLatestFrom not preserving semantics 


tlaitinen:
The following code 
```javascript
import Rx from 'rxjs/Rx';

const s1 = Rx.Observable.interval(1000).take(3);
const s2 = Rx.Observable.interval(100).take(30);
const s3 = s1.withLatestFrom(s2, (v1,v2) => [v1,v2])
s3.subscribe(function (x) { console.log(x) })
```
outputs 
```
[ 0, 8 ]
[ 1, 18 ]
[ 2, 28 ]
```

According to FAQ, `withLatestFrom` can be expressed in xstream as follows:

```javascript
import xs from 'xstream'

const s1 = xs.periodic(1000).take(3);
const s2 = xs.periodic(100).take(30);
const s3 = s1.map((v1) => s2.map((v2) => [v1,v2])).flatten(); 
s3.addListener({
    next: x => console.log(x),
    error: err => console.error(err),
    complete: () => console.log('completed')
})
```

However, it outputs 30 tuples instead of 3:
```
[ 0, 0 ]
[ 0, 1 ]
[ 0, 2 ]
[ 0, 3 ]
[ 0, 4 ]
[ 0, 5 ]
[ 0, 6 ]
[ 0, 7 ]
[ 0, 8 ]
[ 1, 9 ]
[ 1, 10 ]
[ 1, 11 ]
[ 1, 12 ]
[ 1, 13 ]
[ 1, 14 ]
[ 1, 15 ]
[ 1, 16 ]
[ 1, 17 ]
[ 1, 18 ]
[ 2, 19 ]
[ 2, 20 ]
[ 2, 21 ]
[ 2, 22 ]
[ 2, 23 ]
[ 2, 24 ]
[ 2, 25 ]
[ 2, 26 ]
[ 2, 27 ]
[ 2, 28 ]
[ 2, 29 ]
completed
```

ntilwalli:
Proper `withLatestFrom` translation is:
```
// In rx 4
a$.withLatestFrom(b$)
// In xstream
const bMem$ = b$.remember(); 
a$.map(a => bMem$.take(1).map(b => [a, b]).flatten()
```

ntilwalli:
[...]

PR #123, opened 2016-09-23
Implement `sampleCombine` extra

### LINKED ISSUE #102: Implement "withLatestFrom" as an extra operator
I'm looking for a method to accomplish:

```
--1----2-----3--------4---
----a-----b-----c--d------
       combineLeft
--1?---2a----3b-------4d--
```

I believe that it is called `withLatestFrom` in RxJS. I think it is a really common case in cycle.js development.

For now, I'm using:

```
let withLatestFrom = (a, b) => {
	let c = xs.create()
	let bValue = ''
	b.addListener({
		next: v => { bValue = v },
		error: () => {},
		complete: () => {}
	})
	a.addListener({
		next: v => c.shamefullySendNext([v, bValue]),
		error: () => {},
		complete: () => {}
	})
	return c
}
```

1. Is it possible, to easly compose it from existing operators?
2. Maybe it could be implemented as an extra?

xtianjohns:
Add an extra named `sampleCombine`. This extra is designed to provide parity with the [`withLatestFrom`](http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html#instance-method-withLatestFrom) operator, and similar operators, in other stream libraries.

## Description
SampleCombine combines a source stream with multiple other streams. The result stream will emit the latest events from all input streams, but only when the source stream emits.

If the source, or any input stream, throws an error, the result stream will propagate the error. If any input streams end, their final emitted value will remain in the array of any subsequent events from the result stream (this behavior is consistent with the implementation in Rx). The result stream will only complete upon completion of the source stream.

This extra operator implements the CombineSignature core interface.

Fixes #102.
Related to #97.

Additionally, I've included a bug fix for variadic types in `MergeSignature`, and implemented similar expanded types for `CombineSignature`. Neither of those changes are required for the extra to function.

## Motivation and Context
`withLatestFrom` has been a tricky operator to explain and compose for many `xstream` users, as demonstrated in #102 and #97.

This extra aims to make implementing the kind of functionality achievable with `withLatestFrom` straightforward. No selector support is introduced here, but users can select items using `filter` or `map` on a result stream.

## How Has This Been Tested?
My implementation for the operator borrowed heavily from `CombineProducer` and implements `CombineSignature`. I've implemented all of the same tests for `combine` in `sampleCombine`.

The `should not break future listeners` test for `combine` may not actually be relevant for this extra, but I included it anyway. I believe that the test is supposed to verify that a teardown step in a producer does not create errors later in a combine lifecycle. I think my test verifies this, but I can make changes if needed.

## Checklist:

### DIFF
--- a/markdown/footer.md
+++ b/markdown/footer.md
@@ -16,12 +16,14 @@ A: Read this [blog post](http://staltz.com/why-we-built-xstream.html) on the top
 
 **Q: What is the equivalent of [`withLatestFrom`](http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html#instance-method-withLatestFrom) in xstream?**
 
+`withLatestFrom` is implemented as an extra named [`sampleCombine`](https://github.com/staltz/xstream/blob/master/EXTRA_DOCS.md#sampleCombine). It may also be composed with existing operators:
+
 <!-- skip-example -->
 ```js
 A.withLatestFrom(B, (a, b) => a + b)
 ```
 
-can be achieved in *xstream* with
+can be composed in *xstream* with
 
 ```js
 B.map(b => A.map(a => a + b)).flatten()
--- a/src/core.ts
+++ b/src/core.ts
@@ -227,7 +227,7 @@ export interface MergeSignature {
     s5: Stream<T5>,
     s6: Stream<T6>,
     s7: Stream<T7>,
-    s8: Stream<T7>): Stream<T1 | T2 | T3 | T4 | T5 | T6 | T7 | T8>;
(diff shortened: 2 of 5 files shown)


staltz:
@xtianjohns FYI I also added this test and change: https://github.com/staltz/xstream/commit/9882e8950073a8f2ae8fd61fe83d46b1ccd25cbf#diff-34882d4da95c7cb8346862063f9e58c8R60

To align with RxJS, and align with xstream's combine.

xtianjohns:
Thanks, @staltz. I much prefer the `compose` API, your approach (more like dropUntil) in `6.4.0` is a big improvement.

giphy

PR #126, opened 2016-10-03
improve TypeScript typings for delay, dropRepeats, dropUnti, split

whitecolor:
A little bit breaking because `delay` doesn't not allow to specify type, so one will have to fix it in the codebase.


staltz:
Thanks @whitecolor (even though I took a long time to merge it)

whitecolor:
@staltz some conflicts there were?

staltz:
Should have released it as major (breaking change).

whitecolor:
Ah ok, I was on vacation anyway, so I didn't miss much 😄

PR #131, opened 2016-10-18
docs(README): document throttle operator

geovanisouza92:
I noticed that throttle operator wasn't on EXTRA_DOCS.md

### DIFF
--- a/EXTRA_DOCS.md
+++ b/EXTRA_DOCS.md
@@ -14,6 +14,7 @@
 - [`pairwise`](#pairwise) (operator)
 - [`sampleCombine`](#sampleCombine) (operator)
 - [`split`](#split) (operator)
+- [`throttle`](#throttle) (operator)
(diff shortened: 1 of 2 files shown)


xtianjohns:
Hey @geovanisouza92, this document will be generated automatically during the build process. If there's something missing from the documentation that is hosted, you can work on the relevant documentation templates in `markdown/` or directly in source.

Does that make sense?

geovanisouza92:
@xtianjohns thanks for reply. I did run `npm run extra-docs` but it doesn't updated the EXTRA_DOCS.md. So I thought that was a bug, as staltz have released 6.5.0 an the site still doesn't mention throttle extra.

Feel free to close the PR if I have understood it wrong. I'm just trying to help. :)

staltz:
Thanks @geovanisouza92 but indeed this is something automatic. Should be fixed now/soon

PR #136, opened 2016-10-24
Update listener to only require next:

SteveALee:
This time the change is made to the correct source file


staltz:
Thanks @SteveALee :)

Issue #140, opened 2016-11-07
Make create api more convinient (like in Rx)


whitecolor:
I think it is reasonable to improve (simplify) `create` interface (to make it compatible with rx).

Current `create` stream api:
```ts
xs.create({
  start: () => {}
  stope: () => {}
})
```
looks original and explicit but not convenient in many cases. Because it forces you to use some context to share objects between start and stop handler and eventually it looks quite clumsy, esp. when you want to be typesafe.

This is the example, instead of this:
```ts
interface Context {
  job: Job
}
xs.create({
  // should also use not arrow function, to have a context
  start: function (this: Context, observer) => {
    ...
    this.job = new Job()
    ..
  }
  stop: function(this: Context) {
    this.job.stop()
  }
})
```

I would rather have this:
```ts
xs.create((observer) => {
    ...
    let job = new Job()
    ...
    return () => job.stop()
})
```

This would be a logical step after such improvement was implemented with `addListener` api.

staltz:
There are two possible strategies here:

1. Move Producer/start/stop architecture entirely to the Observable/subscription architecture
2. Add support for an additional API like `xs.create((observer) => ... return unsubscribe)`

Option (1) is too huge, would probably bring tons of breaking changes. Probably not feasible.

Option (2) is actually already supported, indirectly. We have `fromObservable`. So you can do:

```js
xs.fromObservable(Rx.Observable.create((observer) => {
  // ... 
  return unsubscribe;
}));
```

So, closing this issue.

whitecolor:
The proposal was just about simplifying `xs.create` API, the Option (2) is not viable for this case, one wouldn't want to use Rx dependency just to be able to use such API.

What is the problem with supporting Option (1) along with current Produceer API?

Here
https://github.com/staltz/xstream/blob/master/src/core.ts#L73

producer could be initialized if it is object with `start` and `stop`, or it is a `start` function, that returns `stop` (if there will not be the problem with sync streams that stops just as it starts)

staltz:
How would you set the `_stop` in internalizeProducer? Try to make a PR and you'll see the problems with doing this. Also, it would add more code to core.

whitecolor:
Yes there are problems with sync streams.

whitecolor:
I've made a pull, it should work https://github.com/staltz/xstream/pull/151

PR #151, opened 2016-11-28
feat(create): Allows to create stream using producer function

whitecolor:
Allows to create stream using producer function that returns dispose function

Should be returned stop/dispose function required?

#140

### DIFF
--- a/src/core.ts
+++ b/src/core.ts
@@ -51,6 +51,10 @@ export interface Producer<T> {
   stop: () => void;
 }
 
+export interface ProducerFn<T> {
+  (listener: Listener<T>): () => void;
+}
+
 export interface Listener<T> {
   next: (x: T) => void;
   error: (err: any) => void;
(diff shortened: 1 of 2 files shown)


tests/stream.ts
```diff
@@ -74,6 +74,33 @@ describe('Stream', () => {
...
+    const expected = [10, 20, 30];
+    let producerStopped: boolean = false;
+
+    const stream: Stream<number> = xs.create<number>((listener) => {
+      listener.next(10);
+      listener.next(20);
+      listener.next(30);
+      listener.complete();
+      return () => {
+        done();
```
TylorS: done() needs to be called *after* assertions.

src/core.ts
```diff
@@ -70,15 +74,32 @@ export type Observable<T> = {
...
-function internalizeProducer<T>(producer: Producer<T>) {
+function internalizeProducer<T>(producer: Producer<T> | ProducerFn<T>) {
   (producer as InternalProducer<T> & Producer<T>)._start =
     function _start(il: InternalListener<T>) {
       (il as InternalListener<T> & Listener<T>).next = il._n;
       (il as InternalListener<T> & Listener<T>).error = il._e;
-      (il as InternalListener<T> & Listener<T>).complete = il._c;
-      this.start(il as InternalListener<T> & Listener<T>);
+      if (typeof producer === 'function') {
+        let _stoped = false;
```
TylorS: Spelling mistake. `_stoped` -> `stopped`

src/core.ts
```diff
@@ -70,15 +74,32 @@ export type Observable<T> = {
 export type FromInput<T> = Promise<T> | Stream<T> | Array<T> | Observable<T>;
 
 // mutates the input
-function internalizeProducer<T>(producer: Producer<T>) {
+function internalizeProducer<T>(producer: Producer<T> | ProducerFn<T>) {
```
TylorS: My personal feeling on this is that it should be a decision on either an object `{ start, stop }` *or* `(listener) => disposeFn`. Having both is a complication to the overall design IMO.

tests/stream.ts
whitecolor: This is completely taken from upper `create` test, and it passes of course.

whitecolor: Yes this is just a proposal, and I showed that it can be implemented if needed. I think it brings not much complication, but allows to align with standard observable approach.

src/core.ts
```diff
@@ -70,15 +74,32 @@ export type Observable<T> = {
...
       (il as InternalListener<T> & Listener<T>).next = il._n;
       (il as InternalListener<T> & Listener<T>).error = il._e;
-      (il as InternalListener<T> & Listener<T>).complete = il._c;
-      this.start(il as InternalListener<T> & Listener<T>);
+      if (typeof producer === 'function') {
+        let _stoped = false;
+        (il as InternalListener<T> & Listener<T>).complete = function _stop () {
+          (producer as InternalProducer<T> & ProducerFn<T>)._stop
+            ? (producer as InternalProducer<T> & ProducerFn<T>)._stop()
+            : _stoped = true;
```
staltz: Performance problem: `_stoped` is from the closure, and closures are expensive.

src/core.ts
```diff
@@ -70,15 +74,32 @@ export type Observable<T> = {
...
+function internalizeProducer<T>(producer: Producer<T> | ProducerFn<T>) {
   (producer as InternalProducer<T> & Producer<T>)._start =
     function _start(il: InternalListener<T>) {
       (il as InternalListener<T> & Listener<T>).next = il._n;
       (il as InternalListener<T> & Listener<T>).error = il._e;
-      (il as InternalListener<T> & Listener<T>).complete = il._c;
-      this.start(il as InternalListener<T> & Listener<T>);
+      if (typeof producer === 'function') {
+        let _stoped = false;
+        (il as InternalListener<T> & Listener<T>).complete = function _stop () {
```
staltz: Performance problem: since the function `_stop` is created inline here, every time this producer starts, a new object allocation happens.

staltz:
I put some comments inline. Basically if we were to remove the closures and dynamic allocations we would have to introduce some new classes, and that would significantly add to the size of core. I think like Tylor here, that it should be either `{start, stop}` or `(listener) => dispose`, but not both. I'm not claiming start/stop is the best, but migrating to `(listener) => dispose` is too much of a radical rewrite now.

staltz:
This is an ancient PR, so I hope closing it is okay. The concern is real, though, and I think the `addListener`/`removeListener` API was an experiment to see how people understand it as being different to `subscribe`/`unsubscribe` API, but I believe the add/remove API is not so nice, and if I would rewrite xstream, I would do something like `subscribe` returning Subscription, or a callbag style of unsubscribing.

Issue #153, opened 2016-11-30
Stream.combine().pairwise() sends the same value for prev and next


appsforartists:
[reduction](http://codepen.io/appsforartists/pen/KNyqKr?editors=0010)

```javascript
Stream.combine(a$, b$).compose(pairwise).map(
  ([[prevA, prevB], [nextA, nextB]]) => //...
);
```

doesn't work as expected because `prev` and `next` share a reference.  You can work around this by applying pairwise to each stream before combining them (and adjusting the predicate's signature accordingly):

```javascript
Stream.combine(
  a$.compose(pairwise), 
  b$.compose(pairwise)
).map(
  ([[prevA, nextA], [prevB, nextB]]) => //...
);
```

or adding a map to the chain:

```javascript
Stream.combine(a$, b$).map(pair => [...pair]).map
```

staltz:
I'm aware why this happens. The output of combine is always the same array reference. This is done for performance. In one perspective it's a feature (faster performance), in another perspective it's a bug (this issue). I'm not sure yet what to do but I'm preferring to keep the current behavior.

Curious to hear more thoughts. 

(PS: this issue is basically the same as https://github.com/ReactiveX/rxjs/issues/1686)

appsforartists:
Once I realized it was recycling the reference, I understood what was happening and a workaround became trivial. 

I haven't tested if pairwise also pools references, but it probably should for the same reason. 

Perhaps a note in the combine docs is the way forward:

> **Note:** combine uses the same array instance for each iteration to minimize garbage collection.  If you need to compare emissions over time, either apply pairwise to the streams before passing them to combine, or map the combine output through [].slice. 

pairwise could also test if prev and next are objects and identical references, and warn if they are. 

I can open a PR with that warning and docs note, linking to gists for the workarounds?  It's a similar solution to the MemoryStream/imitate bug.

staltz:
I think a PR to add a note in the docs for combine would be nice.

jsdevtom:
I made [`distinctUntilChangedImmutable`](https://github.com/jsdevtom/distinct-until-changed-immutable) just for this

PR #155, opened 2016-11-30
Added note about xs.combine recycling its array instance

appsforartists:
Closes https://github.com/staltz/xstream/issues/153

PR #156, opened 2016-12-02
docs(combine): add note about recycling array instance across emissions

appsforartists:
Closes https://github.com/staltz/xstream/issues/153

### DIFF
--- a/src/core.ts
+++ b/src/core.ts
@@ -1618,6 +1618,16 @@ export class Stream<T> implements InternalListener<T> {
    * be emitted on the output stream. It's essentially a way of joining together
    * the events from multiple streams.
    *
+   * Note: to minimize garbage collection, *combine* uses the same array
+   * instance for each emission.  If you need to compare emissions over time,
+   * cache the values with `map` first:
+   *
+   * ```js
+   * xs.combine(stream1, stream2).map(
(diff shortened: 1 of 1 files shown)


appsforartists:
Took the same note from https://github.com/staltz/xstream/pull/155 and moved it to JSDoc

src/core.ts
```diff
@@ -1618,6 +1618,15 @@ export class Stream<T> implements InternalListener<T> {
...
    *
+   * Note: to minimize garbage collection, *combine* uses the same array
+   * instance for each emission.  If you need to compare emissions over time,
+   * cache the values with `map` first:
+   *
+   * ```js
+   * xs.combine(stream1, stream2).map(
+   *   combinedEmissions => ([ ...combinedEmissions ])
+   * ).compose(pairwise)...
+   *
```
staltz: Missing closing triple quotes here.

Widdershin: Also the trailing `...` on line 1628 will cause `markdown-doctest` to fail. Could it be removed?

appsforartists: Good catch.  Did this a bit too quickly.

Thanks!

staltz:
Thanks

PR #159, opened 2016-12-17
bug: unterminated take operator

### LINKED ISSUE #158: Circular dependency with take(1) reaches Maximum call stack size
The following code is able to reach maximum call stack size and I do not see why.

```
var value$proxy = xs.create(); 
var content$ = value$proxy.startWith(1);
var value$ = xs.of(1).map(_ => content$.take(1)).flatten();
value$proxy.imitate(value$)
content$.addListener({
    next: (x) => bin.log("content$proxy: ", x),
  complete: () => bin.log("COMPLETE"),
  error: () => bin.log("ERROR")
})
value$proxy.addListener({
  next: (x) => bin.log("Value$proxy: ", x),
  complete: () => bin.log("COMPLETE"),
  error: () => bin.log("ERROR")
})
```

See also the log of the following webpackbin: http://www.webpackbin.com/4JfBs1HQf 
(**Warning**:  It might take a while to load and you probably don't want to save anything at the risk of crashing your browser tab)

Should not the take(1) operator prevent the proxystream from reaching maximum call stack size?

I believe this construction worked in xstream 6.0.0

MidnightWonderer:
This fixes #158 but IMO `imitate` still not a good interface, though.  
It is very special operator distinct with other operators.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -74,7 +74,7 @@ export interface Subscription {
 
 export interface Observable<T> {
   subscribe(listener: Listener<T>): Subscription;
-};
+}
 
 // mutates the input
 function internalizeProducer<T>(producer: Producer<T>) {
@@ -1182,9 +1182,10 @@ class Take<T> implements Operator<T, T> {
   _n(t: T) {
     const u = this.out;
     if (u === NO) return;
-    if (this.taken++ < this.max - 1) {
+    const m = ++this.taken;
+    if (m < this.max) {
       u._n(t);
-    } else {
+    } else if (m === this.max) {
       u._n(t);
       u._c();
     }
--- a/tests/operator/take.ts
+++ b/tests/operator/take.ts
@@ -79,4 +79,40 @@ describe('Stream.prototype.take', () => {
       },
     });
   });
+
+  it('should terminate properly when "next" function recursively calls itself', (done: any) => {
+    const producer = {
+      start: (listener: any) => {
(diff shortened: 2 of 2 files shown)


staltz:
This looks correct, thanks, but I need to do some modifications to `master` branch so we use TypeScript 2.0. Will merge it soon

staltz:
Can I ask you to rebase this with `master` so that it has only the fix to TakeOperator and the test?

MidnightWonderer:
Done, no conflict.
I don't mind nitpicking comment.
Feel free to review the code style.

src/core.ts
```diff
@@ -1177,9 +1177,10 @@ export class TakeOperator<T> implements Operator<T, T> {
   _n(t: T) {
     const u = this.out;
     if (u === NO) return;
-    if (this.taken++ < this.max - 1) {
+    const taken = ++this.taken;
+    if (taken < this.max) {
       u._n(t);
-    } else {
+    } else if (taken == this.max) {
```
staltz: Can you use triple equals here?

staltz:
Ok, thanks! Can I ask you also to amend the commits so that the scope is `take` and not `TakeOperator`? This will affect what is shown in the changelog.

MidnightWonderer:
I saw massive changes there.  
So I simply do copy/paste work here.

staltz:
Thanks @MidnightWonderer !

PR #164, opened 2017-01-11
Extracted factory interfaces (signatures) from main index.ts.

andersonleite:
I would like to help on extracting some possible dependencies on the index.ts file.

For a first commit, I suggest to extract the two factories interfaces (signatures): MergeSignature and  CombineSignature.

Since we only have two, I think the best approach is to have a signatures directory inside factory, like this:
src/factory/signature
-combineSignature
-mergeSignature

I used camelCase on the file naming and factory/signature on singular as it's being used as a pattern on the other files/directories.

PS: I prefer the organization above rather then something like:
factory/combine/combineSignature.ts
factory/merge/mergeSignature.ts

The reason in my opinion is that we can have less imports (only one in index.ts) and we only have these signatures.

Let me know if I can keep refactoring to help with PR's like this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,6 +29,7 @@ index.d.ts
 index.js
 index.js.map
 /extra/
+/factory/
 
 # Generated directories
 .sass-cache/
--- a/src/factory/signature/combineSignature.ts
+++ b/src/factory/signature/combineSignature.ts
@@ -1,0 +1,70 @@
+import {Stream} from '../../index';
+
+export interface CombineSignature {
+  (): Stream<Array<any>>;
+  <T1>(s1: Stream<T1>): Stream<[T1]>;
+  <T1, T2>(
+    s1: Stream<T1>,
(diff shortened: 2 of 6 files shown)


staltz:
Hi Anderson, thank you for taking your time to send us this issue. I should make this more clear in the contributing guides, but unfortunately xstream doesn't need refactoring of the code in this style. We put everything in one file so that when browserified there is the least amount of boilerplate code as possible. I know TypeScript signatures like these wouldn't go to the browserified bundle, but anyway it's good not to deviate from the norm.

xstream being carefully engineered to be so tiny makes it hard to modify or contribute to. If you want to help, I'd recommend looking at existing issues with high priority or marked as "help wanted".

Better even, I recommend taking a look at the cyclejs repo issues, there are many more there that would appreciate some helping hand.

Hope you understand if I close this PR and I'll update the contributing docs to reflect these decisions.

PR #172, opened 2017-02-17
Reject method

joaomilho:
I got a little tired of doing...  `.filter(R.complement(R.isNil))`.

I'm not really sure on the implementation 'though. If you wanna keep it, LMK if it's ok.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1745,6 +1745,41 @@ export class Stream<T> implements InternalListener<T> {
     return new Stream<T>(new Filter<T>(passes, this));
   }
 
+  reject<S extends T>(passes: (t: T) => t is S): Stream<S>;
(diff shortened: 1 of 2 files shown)


staltz:
Hi @joaomilho, thanks for the work you put into this PR. On the other hand, we probably don't take anymore new simple operators, due to the design goals in contributing.md: https://github.com/staltz/xstream/blob/master/CONTRIBUTING.md

joaomilho:
Fine, how do you suggest this should be done then? extra/other lib?

staltz:
`R.complement` or `filter(x => x != null)`. Consider how the addition of `reject` would make it quicker to *write* code, but slower to *read and understand* for other developers, because they need to read up about yet another operator.

joaomilho:
Fair enough.

Issue #176, opened 2017-03-28
Replace reference path definition with reference types in fromEvent.ts 


Steelfish:
Currently the fromEvent.ts file references a local node type definition. This causes conflicts when you are locally using the node type definition. The line: 
`/// <reference path="../../typings/globals/node/index.d.ts" />` 
should be replaced with:
 `/// <reference types="node" />`

See also:  https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html

PR #180, opened 2017-04-07
feat: Moved to @types declaration files

Steelfish:
Instead of using typings, xstream now uses the @types packages for declaration files as recommended
for typescript 2.0+. This also ensures compatibility with projects that used these declaration
files. See: https://blogs.msdn.microsoft.com/typescript/2016/06/15/the-future-of-declaration-files/

This fixes https://github.com/staltz/xstream/issues/176 and compilation errors for projects using the @types declaration files such as seen in: https://gist.github.com/EnigmaCurry/c4849b25484d5068936a6818038146df


staltz:
Okay, this looks great. Thanks @Steelfish

PR #187, opened 2017-04-24
Relax promise typing

MidnightWonderer:
This corresponds to  
https://github.com/ReactiveX/rxjs/pull/2505  
Promise definition is not compatible with Promise implementations out there.  
I think the type should be relaxed.  
I can add more test like what I did in rxjs but I think the type itself should suffice.  
Poke me if you think the test is necessary.


staltz:
Looks good. Thanks!

PR #189, opened 2017-05-01
Remove exessive stuff from npm publication

whitecolor:
.sass-cache, typings ... more take than 500k, half of the package, this is useful to reduces this size when bundling xstream to some extensions, uploading in containers etc. Any way it is good when published package only has what its need and size is lean.

`dist` should stay as may be needed by some bundlers.


whitecolor:
Actually it is better to update `files` entry in package.json not do this verbose listing in .npmignore

staltz:
I'm okay with this list in .npmignore
Thanks @whitecolor !

PR #193, opened 2017-05-10
attempt to improve Stream doc section

mightyiam:
During #goodnessSquad.

### DIFF
--- a/markdown/overview.md
+++ b/markdown/overview.md
@@ -4,12 +4,19 @@ XStream has four fundamental types: Stream, Listener, Producer, and MemoryStream
 
 ## Stream
 
-A Stream is an **event emitter** with multiple Listeners. When an event happens on the
-Stream, it is broadcast to all its Listeners at the same time.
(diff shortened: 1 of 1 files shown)


markdown/overview.md
```diff
@@ -4,12 +4,19 @@ XStream has four fundamental types: Stream, Listener, Producer, and MemoryStream
 
 ## Stream
 
-A Stream is an **event emitter** with multiple Listeners. When an event happens on the
-Stream, it is broadcast to all its Listeners at the same time.
+A Stream is an *event emitter*. It can have any number of *listeners*. When an event occurs on the
+Stream, it is broadcast to all its *listeners*, synchronously and sequentially.
```
mightyiam: Is it true that it is synchronously and sequentially? Perhaps the term sequentially is redundant?

staltz:
Hi @mightyiam , thank you for taking your time to send the PR. I appreciate it, however this is just a paraphrasing of the existing docs. It's not adding value, and the language used is just a matter of style. I think we are not lacking these types of PRs, but we still have a couple of issues marked "help wanted" in the repo.

Issue #142, opened 2016-11-12
buffer operator or toArray


jvanbruegge:
It would be very helpful to be able to accumulate all values held back by debounce in an array (like in RxJS). That way, you can implement triple clicks or similar behavior far easier.

xtianjohns:
Hey @SuperManitu, I'm not sure of your exact use case with debounce, but we should be able to implement aggregation and buffering by composing several xstream core and extra operators.

For example, if you needed debounce to "buffer" values and emit them as a group (like RxJS `buffer`), you can use xstream's [`split`](https://github.com/staltz/xstream/blob/master/EXTRA_DOCS.md#split) extra. It behaves very similarly to Rx's [`window`](https://github.com/Reactive-Extensions/RxJS/blob/master/doc/api/core/operators/window.md) operator.

Here is an example of using it to split click events into counts that occur within a time window:

```ts
import xs from 'xstream';
import flattenSequentially from 'xstream/extra/flattenSequentially';
import fromEvent from 'xstream/extra/fromEvent';
import split from 'xstream/extra/split';

/* Get the button. */
const button = document.getElementById( 'button' );

/* Create a .5 second timer. */
const halfSecond = xs.periodic( 500 );

/* Capture all click events on a button. */
const allClicks = fromEvent( button, 'click' );

// Group events in the click stream.
const counts = allClicks
  .compose( split( halfSecond ) )
  .map( stream => stream.fold( count => count + 1, 0 ).last() )
  .compose( flattenSequentially );
```
Then, for a triple click implementation (which you mentioned), you could filter these counts.
```ts
const triples = counts.filter( clicks => clicks > 2 ); // triple click happened
```
Now, arguably, this doesn't get you to a true "triple-click" because you could click three times during an overlap between two half-second periods. But an alternative approach can fix that:
```ts
// Emit if 3 clicks happen in .5 seconds.
const triples = buffer( allClicks, 3, halfSecond );

function buffer( source, count, closing ) {
  const grouped = source.map( item => {
    // For every click, track the next two clicks.
    const buffered = source.take( count - 1 );
    // End the stream if 'closing' emits first.
    return buffered.endWhen( closing ).last();
  });
  return grouped.compose( flattenSequentially );
}
```
So, depending on one's needs, composing operators can create some really powerful functionality.

If either of these fit the use cases you had in mind, do you still think it's beneficial to implement a change to `debounce` or add an extra `buffer` operator?

[...]

PR #196, opened 2017-05-11
feat(buffer): buffer operator that splits input stream into arrays by …

VKobeliatsky:
…separator stream

buffer operator accepts a separator stream as an argument. It accumulates input stream events into
arrays and emits them when separator stream emits or closes. Output stream is closed when separator stream is closed. #goodnessSquad

### DIFF
--- a/src/extra/buffer.ts
+++ b/src/extra/buffer.ts
@@ -1,0 +1,122 @@
+import {Operator, Stream, Subscription, InternalListener, OutSender, NO_IL} from '../index';
+
+class SeparatorIL<T> implements InternalListener<any>, OutSender<Array<T>> {
+    constructor(public out: Stream<Array<T>>,
+                private op: BufferOperator<T>) {
+    }
+
+    _n(t: any) {
+        this.op.flush();
+    }
+
+    _e(err: any) {
+        this.out._e(err);
+    }
+
+    _c() {
+        this.op.flush();
+        this.out._c();
+    }
+}
+
+class BufferOperator<T> implements Operator<T, Array<T>> {
+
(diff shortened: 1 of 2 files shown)


VKobeliatsky:
With @mightyiam during #goodnessSquad

staltz:
Thanks @VKobeliatsky. This relates to #142 and seems mostly good to merge. I'll do some refactoring after merging, to make the code style consistent with the rest. And also semantically I need to change the test `it('should complete when separator is completed after the source'` because the output stream **should** complete if the **source** completes. It just makes more sense (no reason to keep something open if you know no more values will come) and also RxJS does like this.

VKobeliatsky:
Hi @staltz, thanks for the merge!
The only thing that I am worried about is your suggestion on *complete* behavior. I imagined `separator` stream as a control mechanism over when the `out` stream gets data. So when `separator` completes `out` will never get data and is good to complete, too. If we complete when `input` stream completes we will have to either flush aggregated items immediately and complete (which will break the concept that `separator` is control), or just complete and potentially lose some items.

staltz:
> If we complete when input stream completes we will have to either flush aggregated items immediately and complete (which will break the concept that separator is control)

This is the change that I did. I think it's sensible because these semantics have been used in RxJS for years and people haven't complained about it as far as I know.

VKobeliatsky:
Convinced then, thanks!

Also, looks like I forgot to edit `tsconfig.json`

staltz:
I'll do that, no worries.

mightyiam:
Awesome, @VKobeliatsky!

PR #197, opened 2017-05-15
feat(extra(uniqueBy)): a stream of unique items

VKobeliatsky:
Given a function returns a stream that emits only unique items of input stream with uniqueness
determined by the function

### DIFF
--- a/EXTRA_DOCS.md
+++ b/EXTRA_DOCS.md
@@ -16,7 +16,7 @@
 - [`split`](#split) (operator)
 - [`throttle`](#throttle) (operator)
 - [`tween`](#tween) (factory)
-
+- [`uniqueBy`](#uniqueBy) (operator)
 # How to use extras
 
 The following are standalone stream operators and stream factories that may be separately imported and utilized in your project.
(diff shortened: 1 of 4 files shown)


mightyiam:
Go Vova!

staltz:
Hi @VKobeliatsky, this is very decently made PR and good operator, but I don't want there to be too many extra operators. uniqueBy is like RxJS distinct, and it's one of the more rarely used operators. You can easily take this same code you made and publish it as a library on npm, then using it will be in the same style of import+compose. If you want, you can also edit this repo's Wiki to mention it there. If you're interested in other xstream-specific developments, it's usually a good idea to open an issue beforehand, to discuss the viability of the code change. Often the contributing.md is useful too: "It must have few operators (either core **or extra**) to reduce decision paralysis"

VKobeliatsky:
Fair enough, thanks.

mightyiam:
@VKobeliatsky please do make an independent package. I'm available for 1:1 on any questions you may have on the process, gratis.

PR #203, opened 2017-07-03
Fix typo "fo each" => "for each"

chrischrischris:


staltz:
Hi @chrischrischris, it would be nice to edit the source code with the JSDoc instead, because the README is automatically generated from those. If it's too much effort for a PR, I can do it myself. Thanks anyway

chrischrischris:
No problem will submit that soon, I didn't realize it was auto generated.

Opened https://github.com/staltz/xstream/pull/204

Thanks for this and cycle.js!  It's had me go all in with FRP.  Now if I can only convince my work to use it ...

PR #206, opened 2017-07-12
fix(extra): Simplify extra operator generics

### LINKED ISSUE #202: Extra Operator Generics in Typescript 2.4
Typescript 2.4 has better generic type inference, including better support for unifying return types in generic functions with the generic type parameter. I'm getting type inferencing issues in Typescript 2.4 that I can't seem to solve (x is not x errors) trying to use some of the extra operators and I think it's because of the way the generics are used in the library.

Take the signature for `dropRepeats` for example:

```ts
export default function dropRepeats(isEqual: (<T>(x: T, y: T) => boolean) | undefined = void 0): <T>(ins: Stream<T>) => Stream<T>
```

In Typescript 2.4 I'm getting an error because the generic parameter `T` for the `isEqual` is not the generic parameter `T` for the return function. That's not something that Typescript prior to 2.4 has checked.

It seems to me that the obvious solution to move the generic parameter to the outer function would both fix this problem for Typescript 2.4 and happen to give better type inferencing in Typescript before 2.4:

```ts
export default function dropRepeats<T>(isEqual: ((x: T, y: T) => boolean) | undefined = void 0): (ins: Stream<T>) => Stream<T> {
  return function dropRepeatsOperator(ins: Stream<T>): Stream<T> {
    return new Stream<T>(new DropRepeatsOperator<T>(ins, isEqual));
  };
}
```

This updated version of the operator function passes all the existing tests in Typescript 2.1 (current package.json version). The tests don't currently run in Typescript 2.4; seems like there are updates need to be made to the mocha and other types first.

I'm happy to make a PR to move the generics in the extra operators to this simpler form, but I wanted to make sure I wasn't missing some reason for why the generics were written the way they were originally.

WorldMaker:
Move the generic parameters of the extra operators to support better generic parameter inferencing in Typescript 2.4 while continuing to pass tests in Typescript 2.1. Delay is the only extra operator of this form skipped in this commit because it is the only one that didn't pass Typescript inferencing compile (in imitate tests) in Typescript 2.1 (though it builds in Typescript 2.4).

Resolves #202

### DIFF
--- a/src/extra/buffer.ts
+++ b/src/extra/buffer.ts
@@ -104,8 +104,8 @@ class BufferOperator<T> implements Operator<T, Array<T>> {
  * split the output stream.
  * @return {Stream}
  */
-export default function buffer(s: Stream<any>): <T>(ins: Stream<T>) => Stream<Array<T>> {
-  return function bufferOperator<T>(ins: Stream<T>) {
+export default function buffer<T>(s: Stream<any>): (ins: Stream<T>) => Stream<Array<T>> {
+  return function bufferOperator(ins: Stream<T>) {
     return new Stream<Array<T>>(new BufferOperator<T>(s, ins));
   };
 }
--- a/src/extra/debounce.ts
+++ b/src/extra/debounce.ts
@@ -90,8 +90,8 @@ class DebounceOperator<T> implements Operator<T, T> {
  * @param {number} period The amount of silence required in milliseconds.
  * @return {Stream}
  */
-export default function debounce(period: number): <T>(ins: Stream<T>) => Stream<T> {
-  return function debounceOperator<T>(ins: Stream<T>) {
+export default function debounce<T>(period: number): (ins: Stream<T>) => Stream<T> {
+  return function debounceOperator(ins: Stream<T>) {
     return new Stream<T>(new DebounceOperator(period, ins));
   };
 }
--- a/src/extra/dropRepeats.ts
+++ b/src/extra/dropRepeats.ts
@@ -112,8 +112,8 @@ export class DropRepeatsOperator<T> implements Operator<T, T> {
  * checks if it is equal to previous event, by returning a boolean.
  * @return {Stream}
  */
-export default function dropRepeats(isEqual: (<T>(x: T, y: T) => boolean) | undefined = void 0): <T>(ins: Stream<T>) => Stream<T> {
-  return function dropRepeatsOperator<T>(ins: Stream<T>): Stream<T> {
+export default function dropRepeats<T>(isEqual: ((x: T, y: T) => boolean) | undefined = void 0): (ins: Stream<T>) => Stream<T> {
+  return function dropRepeatsOperator(ins: Stream<T>): Stream<T> {
     return new Stream<T>(new DropRepeatsOperator<T>(ins, isEqual));
   };
 }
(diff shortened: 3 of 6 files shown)


staltz:
It's odd that the delay-imitate case happens, I don't understand why. By the way, wouldn't you say this is a `refactor` commit instead of `fix`? I can't see any *change* to my app when applying this.

WorldMaker:
I think that delay-imitate case is one of the ones Typescript 2.4 was exactly targeting to fix/better type inferencing.

It's a `fix` for type information, so from the perspective of Typescript compile-time usage it's a fix. But yes, from the perspective of runtime it has no change/impact. It's an interesting gray area between `fix`/`refactor`.

For what it helps, to work around the type problem in TS 2.4 I had to `any` type a couple parameters as that was the only working option. From my perspective of `--noImplicitAny` and that even explicit `any` are a TODO/FIXME marker, it will be nice to fix that once this gets published I can remove a couple `any` typed parameters. Even better this will fix the behavior in older TS too, giving you some type inference where that was none before and sometimes an implicit `any` or `{}`.

staltz:
I'm curious which case(s) would require a `any` type? I don't understand, as even before, the inferred type after `compose` + extra operator was a non-`{}` type.

WorldMaker:
Here's where I've dealt with it the most:

```ts
let stream: Stream<Position>
stream.compose(dropRepeats((x: Position, y: Position) => false /* insert better comparison here */))
```

With `--noImplicitAny` set, I have to explicitly declare the types of `x` and `y`, it wouldn't infer them as anything better than an implicit `any` (and thus a compile error with `--noImplicitAny`).

In 2.4 I started getting an error that Position is not Position (due to generic inferencing fun) and the only fix I found was to explicit `any`:

```ts
let stream: Stream<Position>
stream.compose(dropRepeats((x: any, y: any) => false /* insert better comparison here */))
```

After this PR, this should work:

```ts
let stream: Stream<Position>
stream.compose(dropRepeats((x, y) => false /* insert better comparison here */))
```

With proper type inferencing that `x` and `y` are `Position`. (Even in 2.1.)

staltz:
Okay, that's a great explanation. Could you add such similar case as a test? Then it makes it clear that this is a fix (the proposed test does not pass without the proposed changes to `src`).

WorldMaker:
Updated a test locally and it looks like I was wrong and 2.1 doesn't catch the inference (it inferred `{}`), so it does need 2.4's return type unification to work.

staltz:
So there is no test that could be written for this PR? Would it then be a refactor and not a fix?

WorldMaker:
You tell me. From my perspective, it doesn't matter what category you want to lump it in, it's your category system.

staltz:
Basically before and after this PR, the proposed test

```typescript
let stream: Stream<Position>
stream.compose(dropRepeats((x, y) => false /* insert better comparison here */))
```

would fail, but for different reasons. Removing one type of error and adding another type of error, for the same use case, seems to me like a refactor. However, there may be a use case which we are not predicting which may break if the previous type of error was ignored with a tsconfig flag, while the new error isn't ignored.

So I'm going to merge this PR in a branch other than master, and start work on updating to TypeScript v2.4, which we can soon publish as xstream v11-rc. We should anyway update to TypeScript v2.4 soon, so I don't see the point of issuing this PR as an xstream v10.x version.

Thanks @WorldMaker

staltz:
I manually merged this, check branch `ts24`

PR #212, opened 2017-08-03
docs(combine): Update combine docs removing Note about array instance

dmorosinotto:
Updated combine docs removing Note and related Comments inside code, becouse according to release 10.3.0 ([e8b2eef](https://github.com/staltz/xstream/commit/e8b2eef)) now it will emit a new Array instance, so the .map trick is no more needed. Added even a test to prove it.


staltz:
Thanks for catching this @dmorosinotto :)

PR #213, opened 2017-08-03
perf(extra/dropRepeats): Optimize set equality check function only inside constructor

dmorosinotto:
Little optimization to extra/dropRepeats to set equality function only inside constructor, not
checking it on every call to isEq, plus some bytes saved in code and deleting an unused public
variable (this.fn) from class.


staltz:
Thanks @dmorosinotto for this improvement. Good PR :)

dmorosinotto:
Happy that you appreciate and merge it

PR #220, opened 2017-10-02
Inline source maps and inline sources

mightyiam:


staltz:
@jvanbruegge if you give an OK to this PR, I'll merge it.

mightyiam:
Hurray no such warnings!

PR #216, opened 2017-09-28
refactor(time): Deprecate time based operators

jvanbruegge:
Move xs.periodic to the extra operators, add a deprecation warning to all time based operators

BREAKING: periodic is no longer part of the `Stream` prototype

to tick another point here: https://github.com/cyclejs/cyclejs/issues/635

PR #222, opened 2017-10-03
refactor(time): Deprecate time based operators

jvanbruegge:
Move xs.periodic to the extra operators, add a deprecation warning to all time based operators

periodic is no longer part of the `Stream` prototype

continuation of #216


staltz:
Great, thanks!

Issue #215, opened 2017-08-20
Add extra operators to Stream


whitecolor:
I think it would be useful to have version of `xs` of `ExtraStream` type that would have methods like `flattenConcurrently`, `sampleCombine` on it defined so they can be used without `compose` semantics.

It could be imported from `import xs, {Stream} from 'xstream/extra'`

Or maybe just add those methods on common `Stream` type, and add them dynamically on Stream prototype when importing `xstream/extra/...`

whitecolor:
@staltz what do you think about it?

xtianjohns:
I think this duplicates #128 a little bit, but in general this sounds like it would be helpful. There were some open questions about TypeScript help when patching, but `rxjs` seems like they've done this at least in part. I know they do module augmentation with `import 'extra/operator';`.

geovanisouza92:
I was to comment exactly that: https://github.com/ReactiveX/rxjs/blob/master/src/add/operator/buffer.ts

`import 'xs/extra/...';` patching prototype seems to be a simple approach from dev experience.

whitecolor:
I know about #128, but the problem with `installing` some operators is just with typing. I'm not aware how Typescript could allow having extended class definitions by importing some modules. 

So I propose to create a separate class that extends common xstream with all the extra methods seem to me as the only normal solution. ES6 tree shaking is here to solve this so "important" front-end bundle size question.

Issue #128, opened 2016-10-06
Idea: `install` helper


staltz:
`install(debounce)` would monkey patch Stream so you could do `a$.debounce(300)`.

xtianjohns:
Would the primary use of the installation helper be to make extras easier to implement for library consumers? If so, would it make sense for the helper to be attached to the extra?
```ts
import xs from 'xstream';
import debounce from 'xstream/extra/debounce';

// install the extra
debounce.install();

// use the extra
xs.of( 'a' ).debounce(300);
```
Alternatively, the function could be exported from core.
```ts
import xs, { install } from 'xstream';
import debounce from 'xstream/extra/debounce';

// install the extra
install( debounce );

// use the extra
xs.of( 'a' ).debounce(300);
```
The former doesn't lend itself to customization (pass your own operators to install). The latter does, and for that reason we'd need to do some name/symbol checking.
```ts
import xs, { install } from 'xstream';

function mapTo( ) { return 1; }

// install the extra
install( mapTo ); // throw here? do nothing?

// okay, it threw. let's give it a name that it likes
function debounce( ) { return 1; }

install( debounce ); // what now?
```

staltz:
> If so, would it make sense for the helper to be attached to the extra?

That's interesting, but ...

> Alternatively, the function could be exported from core.

I prefer this option, because 

> The former doesn't lend itself to customization (pass your own operators to install). The latter does

And

> `install( mapTo ); // throw here? do nothing?`

I think it should do nothing. If we put install in core, it has to be super super small and simple. People doing `install(flatten)` are abusing the API.

One thing I haven't considered is how to support TypeScript. I wish there would be a way, but I'm afraid there won't be. :/

By the way, perhaps `xs.install` is nicer than `import {install} from 'xstream'`.

whitecolor:
What is about typings?

geovanisouza92:
I think I found a way, based on [official docs](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation) using "interface merging" and mixins:

- we create an interface `Stream` and change the current `Stream` to implement `Stream` interface;
- each extra declare an extension to that `Stream` interface and a basic implementation, that is applied as a mixin on `Impl` prototype;

Example:

```typescript
// The Stream interface and core implementation

export interface Stream {
  map(fn: Function): Stream
}

export class Impl {
  map(fn: Function): Stream {
    // ...
  }
}

export function create(producer: object): Stream {
[...]

PR #227, opened 2017-10-29
Embed extra operators to Stream

whitecolor:
This PR embeds extra operators to Stream class.
This PR should contain all operators.

Related issues:
https://github.com/staltz/xstream/issues/128
https://github.com/staltz/xstream/issues/215


staltz:
I appreciate the effort, but as said in #128, I think we won't monkey patch the prototype anymore. I'm happy to keep this (closed) PR around though, so we can reopen and merge if we come to a different decision.

PR #228, opened 2017-10-31
fromEvent is a @factory

vp2177:


staltz:
Hi, did you notice that the `@factory` is already committed in master?

PR #230, opened 2017-11-08
TypeScript@2.6.1 with strictFunctionTypes

mightyiam:
Sorry, this is not a fix. I wasn't sure how prepared I am to tackle an issue of this scope.

Short story is that I don't feel prepared.

So here it is, just a failing config.

My motivation is that it seems that `strictFunctionTypes` requires all dependent code to comply.
And that `strictFunctionTypes: true` is implied by `strict: true`, in v2.6.1.

Now, I don't know the benefits of `strictFunctionTypes`, but I suppose it is encouraged.

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,7 +69,7 @@
     "ts-node": "3.3.x",
     "tsify": "3.0.x",
     "tslint": "5.7.0",
-    "typescript": "2.5.2",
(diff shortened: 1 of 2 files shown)


woutervh-:
My attempt is at #233

staltz:
I suppose it's okay if I close this PR

PR #233, opened 2017-11-24
Use tsconfig/strictFunctionTypes

woutervh-:
Updated Typescript version to 2.6.1, and set the strictFunctionTypes flag to true.

Resulting (type) errors have been fixed.


mightyiam:
This must have been fun.

whitecolor:
This should be carefully reviewed and released as RC, not to break existing code bases.

whitecolor:
BTW flatten typings should be updated to be compatible with cycle HTTP driver (which source returns `Stream<Stream<R> & {request: ..}>`, now typings are:
```ts
  flatten<R>(this: Stream<Stream<R>>): T
```
should be something like that:
```ts
flatten<R, S extends Stream<R>>(this: Stream<S>): Stream<R>
```

staltz:
Hi @woutervh- :)
This PR is decent, and sorry for the very late review. Before merging it, though, test files should remain untouched because we also want to test that the current usage `strictFunctionTypes=false` will still work for people using that setting in their project.

woutervh-:
Hi @staltz thanks for the reply.

As far as I understand it people who use `strictFunctionTypes=false` will always pass any checks that `strictFunctionTypes=true` may fail on. This PR tries also to pass for people using `strictFunctionTypes=true`.

I've run `npm run test` and `npm run compile` with both `strictFunctionTypes=true` and `strictFunctionTypes=false` successfully.

The problem is that TS 2.6.1 will give us type errors with `strictFunctionTypes=true` when we leave the tests unchanged.

We could change back to `strictFunctionTypes=false` and leave the tests as they were, and then users with `strictFunctionTypes=true` will successfully compile. But if new changes are made in future and they break the `strictFunctionTypes=true` check you will only hear about it from issues instead of while compiling. I'm fine with this, I'll leave it up to you :) If you decide to change to `strictFunctionTypes=false` I'll happily revert tests etc.

whitecolor:
> The problem is that TS 2.6.1 

Meanwhile, there is a 2.7.1 that too has some changes, that can affect, and 2.8 is coming

webmaster128:
Could you reconsider merging this. I second this reply from the author: https://github.com/staltz/xstream/pull/233#issuecomment-363129398

staltz:
Note that the current TypeScript dependency in xstream is now 2.8, this PR should revert the changes to the tests and could be updated to bump TypeScript to 2.9 (if it goes smoothly without problems, or keep it as 2.8).

woutervh-:
I have updated the PR to get rid of merge conflicts. Please note that the point of the PR is to set `strictFunctionTypes=true` and doing so will cause tests to fail (by type errors) unless they are changed. That's why that is also included in the PR. If you want to keep `strictFunctionTypes=false` then it defeats the purpose of the PR and we should close it.

staltz:
Hi, we don't have consensus in this discussion that this breaking change for `strictFunctionTypes=true` would be worth it.

webmaster128:
After updating this PR, only diffs in test files are left. This means the rest of the library works for applications that enable strictFunctionTypes.

If the tests (one application using the library) use `strictFunctionTypes` or not is a matter of taste. Enabling it ensures the the library itself does not get regressions. In any case, I don't see how adapting test code is a "breaking change". Users with `strictFunctionTypes = false` can use the library as before. But I guess we can only test one setup if we don't want to create multiple test folders with different typescript settings.

Sinewyk:
@staltz closed via https://github.com/staltz/xstream/pull/268 ?

staltz:
Yes, thank you all for participating in this PR, discussions are as worthwhile as commits. :)

PR #234, opened 2017-12-12
Lock symbol-observable to 1.0.4.

siassaj:
Next version up is 1.1.0 which causes tsc to fail compilation with
error TS2451: Cannot redeclare block-scoped variable 'Symbol'.

Appears to be related to benlesh/symbol-observable#34

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
   },
   "homepage": "https://github.com/staltz/xstream#readme",
   "dependencies": {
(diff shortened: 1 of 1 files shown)


staltz:
Thanks!

benlesh:
@staltz This is fixed in symbol-observable 1.2.0.

benlesh:
I've got a PR at #238  to update it, @staltz. Hope that helps.

PR #242, opened 2018-04-08
Use export default xs to enable autoimport in VSCode

Nitive:
vscode-xs-import-demo

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -2049,4 +2049,5 @@ export class MemoryStream<T> extends Stream<T> {
 }
 
 export {NO, NO_IL};
-export default Stream;
+const xs = Stream
+export default xs;


staltz:
Nice! Thanks

staltz:
Very good point @ryota-ka 
I hadn't intended `xs<T>` to be a valid use of xstream, but since enough people used that, it's a breaking change for them.

staltz:
That said, it's perceivable only for TypeScript users who directly use it. Those who indirectly use it (like in onionify `lib/pickMerge.d.ts`) can reinstall/recompile and no error would happen.

ryota-ka:
@staltz Then any plans to release a new version of `cycle-onionify` with recompiled `lib/*.d.ts`?
A project relying on it reports errors as below and doesn't compile.

```
node_modules/cycle-onionify/lib/pickCombine.d.ts(3,64): error TS2304: Cannot find name 'xs'.
node_modules/cycle-onionify/lib/pickCombine.d.ts(3,95): error TS2304: Cannot find name 'xs'.
node_modules/cycle-onionify/lib/pickMerge.d.ts(3,62): error TS2304: Cannot find name 'xs'.
node_modules/cycle-onionify/lib/pickMerge.d.ts(3,93): error TS2304: Cannot find name 'xs'.
```

staltz:
Thanks! I released onionify v5.1.0 can you test it?

ryota-ka:
@staltz Thank you for your quick response and action.
I tested the new version and everything seems perfect now 🎉

staltz:
@Nitive That's a wise idea. Let's do that, the PR would be welcomed.

PR #245, opened 2018-05-12
use Symbol.observable to get observable

troch:
In fromObservable, use Symbol.observable if present to get observable. If not present,treat received
object as a subscribable (current behaviour).

Some observables passed to `xs.fromObservable` might not have a `subscribe` method. It is also an issue for packages like `redux` or `router5` which have / might have a `subscribe` method with a different interface than observables (accepting listener functions rather than objects).

Note that `xs.fromObservable` accepts subscribables, but `xs.from` doesn't.


staltz:
Useful! Thanks!

PR #246, opened 2018-05-18
add test to check dropRepeats completes

graingert:


staltz:
Very useful PR! Thanks!

tests/extra/dropRepeats.ts
```diff
@@ -21,6 +21,22 @@ describe('dropRepeats (extra)', () => {
       },
     });
   });
+  
+  it('should complete when input completes', (done: any) => {
```
graingert: I was having a bug, thought I'd tracked it down to this, but no it passed on xstream.

PR #253, opened 2018-06-16
Use xstream from npm in website

cprecioso:
GitHub doesn't serve the xstream file with a correct mimetype, so it wouldn't load.
Loading from unpkg.com solves that, and it loads the files from the npm package so it will be kept up-to-date.


staltz:
Thanks!

PR #254, opened 2018-06-16
fix(fromEvent): Improve typings

cprecioso:
`fromEvent` always returned `Stream<any>`. Now it returns a `Stream<Target>` for `EventTarget`s, and else it returns `Stream<any>`, which can be overridden specifying a generic.

```ts
fromEvent(process.stdin, 'data');         // -> Stream<{}>
fromEvent<Buffer>(process.stdin, 'data'); // -> Stream<Buffer>
fromEvent(document.body, 'click');        // -> Stream<Event>
```


src/extra/fromEvent.ts
```diff
@@ -134,12 +134,18 @@ function isEmitter(element: any): boolean {
  * dispatched to any EventTarget beneath it in the DOM tree. Defaults to false.
  * @return {Stream}
  */
-export default function fromEvent(element: EventTarget | EventEmitter,
-                                  eventName: string,
-                                  useCapture: boolean = false): Stream<Event|any> {
+
+function fromEvent<T>(element: EventEmitter, eventName: string): Stream<T>;
```
staltz: Could you use `<T = any>`?

staltz:
Thanks! Good idea :)

src/extra/fromEvent.ts
cprecioso: Maybe it's not necessary since you can already just call the function without specifying the generic and TypeScript will figure it out for you.
But I can add it either way.

staltz: When TypeScript figures it out, then it ignores the `= any`, when it doesn't figure it out, it uses `any` instead of `{}`, which I think is better since `{}` is rarely useful, and the current `fromEvent` outputs `Stream<Event|any>`

cprecioso: Done!

staltz:
Thanks!

PR #258, opened 2018-07-18
fix(dropUntil): align with rxjs skipUntil

### LINKED ISSUE #237: dropUntil doesn't complete output when other completes
Docs of `dropUntil` actually stand that:

> Starts emitting the input stream when another stream emits a next event. The output stream will complete if/when the other stream completes.

But actually, it is not the case, due [to this](https://github.com/staltz/xstream/blob/master/src/extra/dropUntil.ts#L17) `output` stream just start to emit when `other` stream completes.

[Test for this case ](https://github.com/staltz/xstream/blob/master/tests/extra/dropUntil.ts#L28) actually tests that `output` starts when `other` complets, though it is called `should complete the stream when another stream emits complete`.

So there's is something wrong with either the operator's code or its docs/tests.

xtianjohns:
Refactor `dropUntil` to leave its output stream unchanged when its 'other' stream completes. Make output stream empty if 'other' stream is empty.

This change aligns `dropUntil` more closely with the [`skipUntil` operator from rxjs.](https://github.com/ReactiveX/rxjs/blob/37d9ea9a8665d0c4b6593da4c237e037e98f66b5/spec/operators/skipUntil-spec.ts)

## Description

The other internal listener in `dropUntil` now treats completion as a `noop`. Additionally, when source completes, we no longer call `up()`, which seems like a pretty major bug to begin with.

Instead, completion of the source now calls `_stop()`, which removes the internal listener to source and detaches the output stream from the operator. **This is definitely a situation where we need ComVer, since consumers may have unknowingly relied on this behavior, which never removed the internal listener to the source stream.**

Resolves #237.

## Motivation and Context

`dropUntil` is designed to keep a result stream _silent_ until some other stream emits. However, the implementation of `dropUntil` left ambiguous how the operator should behave when either the source stream, or the other (notifier) stream, completes.

The documentation for `dropUntil` gave consumers the impression that the result stream would complete if the 'other' stream completed. However, we had no tests that verified this behavior was correct. Confusion about how the operator should behave led to opening #237.

Aligning the operator's behavior with another implementation (from rxjs) is _okay_, but I consider this change to really be an improvement to the operator. Treating completion as a `noop` is essential because there is no semantic meaning associated with the other stream completing, _for this operator_. If a consumer needed semantics for declaring when their stream should end, they can reach for `endWhen`, or another extra.

Additionally, if the consumer wanted the semantics of `dropUntil` and _didn't_ want `endWhen` behavior, this extra wasn't helpful. They'd have to roll their own.

## How Has This Been Tested?

I've changed one existing test, and added two more. The altered test has had its expectation reversed:
> should _**not**_ complete the stream when another _**non empty**_ stream emits complete

The new tests are largely borrowed from the `rxjs` expectations of `skipUntil`. They can be summarized as:

> If source is infinite, the result stream is infinite.
> If source completes, the result stream completes.

## Checklist:

### DIFF
--- a/src/extra/dropUntil.ts
+++ b/src/extra/dropUntil.ts
@@ -13,9 +13,7 @@ class OtherIL<T> implements InternalListener<any>, OutSender<T> {
     this.out._e(err);
   }
 
-  _c() {
-    this.op.up();
-  }
+  _c() {}
 }
 
 export class DropUntilOperator<T> implements Operator<T, T> {
@@ -63,14 +61,14 @@ export class DropUntilOperator<T> implements Operator<T, T> {
   _c() {
     const u = this.out;
     if (!u) return;
-    this.up();
+    this._stop();
     u._c();
   }
 }
 
 /**
  * Starts emitting the input stream when another stream emits a next event. The
- * output stream will complete if/when the other stream completes.
+ * output stream will emit no items if another stream is empty.
  *
  * Marble diagram:
  *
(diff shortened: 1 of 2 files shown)


staltz:
Really impressive PR, thanks @xtianjohns. No problems found.

It'll cause a breaking change release. (note for self)

PR #259, opened 2018-07-18
feat(debounce): emit pending value on completion

### LINKED ISSUE #127: debounce example (fromDiagram) doesn't work correctly
This example doesn't work correctly, probably problem with `fromDiagram`

``` js
import fromDiagram from 'xstream/extra/fromDiagram'
import debounce from 'xstream/extra/debounce'

const stream = fromDiagram('--1----2--3--4----5|')
 .compose(debounce(60))

stream.addListener({
  next: i => console.log(i),
  error: err => console.error(err),
  complete: () => console.log('completed')
})
```

I see:

```
> 4
> completed
```

Instead of

```
> 1
> 4
> completed
```

### LINKED ISSUE #257: Debounce drops last emitted value if complete emitted too soon
Given this stream:
`const stream = fromDiagram('-1----2-|').debounce(50)`,
1 will be emitted, but not 2, because debounce._c calls this.clearInterval() immediately even if a value is waiting to be emitted.

Debounce drops values with the understanding that a fresher value is coming soon after, but that assumption is violated by dropping the last value.  If debounce's purpose is to wait until a rapidly-changing value has settled down, I propose that when it receives complete, things have obviously settled down and it should emit any pending value immediately and then complete.

I am using debounce to prevent the problem where xs.combine(a, b, c, d) emits three (potentially erroneous) intermediate values if a, b, c, and d all change from the same initial event.  xs.combine(a,b,c,d).debounce(0) should fix the problem nicely, but it doesn't, because of the aforementioned issue.  Another option would be to change combine and merge so that synchronous values from multiple of their inputs only trigger one output emission.

xtianjohns:
Emit any pending value, then complete, if a source stream completes before a debounce interval closes.

## Description
In order to be able to emit the latest value upon completion, the operator needs access to the most recent value. In the previous implementation, the emission took place in a closure created by `setTimeout`, so now the operator has a new property `t` which keeps track of this emission.

This value is cleared when the debounce interval closes and upon completion.

Resolve #257.
Probably should close #127.

### DIFF
--- a/src/extra/debounce.ts
+++ b/src/extra/debounce.ts
@@ -1,10 +1,10 @@
-import {Operator, Stream} from '../index';
+import {Operator, Stream, NO} from '../index';
 
 class DebounceOperator<T> implements Operator<T, T> {
   public type = 'debounce';
   public out: Stream<T> = null as any;
   private id: any = null;
-  private t: any = null;
+  private t: any = NO;
 
   constructor(public dt: number,
               public ins: Stream<T>) {
@@ -37,7 +37,7 @@ class DebounceOperator<T> implements Operator<T, T> {
     this.id = setInterval(() => {
       this.clearInterval();
       u._n(t);
-      this.t = null;
+      this.t = NO;
     }, this.dt);
   }
 
@@ -52,8 +52,8 @@ class DebounceOperator<T> implements Operator<T, T> {
     const u = this.out;
     if (!u) return;
     this.clearInterval();
-    if (this.t != null) u._n(this.t);
-    this.t = null;
+    if (this.t != NO) u._n(this.t);
+    this.t = NO;
     u._c();
   }
 }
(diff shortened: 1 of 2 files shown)


staltz:
Looks good, @xtianjohns, but could you use the special object `NO` instead of `null`? Otherwise a stream that emits nulls (I use these sometimes) could not be correctly debounced.

xtianjohns:
🤦‍♂️ It really has been awhile since I've worked on this - good catch, @staltz.

xtianjohns:
Also, let's hold off on a merge even if CI goes green, Travis failed for `buffer (extra) should not emit empty buffers`, but locally I've had two successes and one failure for `sampleCombine (extra) should complete only when the sample stream has completed`, and never the failure from Travis.

Dunno why I'm seeing flaky behavior, but I'm gonna try to puzzle it out.

staltz:
Thanks!

Issue #161, opened 2016-12-30
Flatten with user-defined concurrency: flattenParallel


plmrry:
There currently is no middle ground between `flattenConcurrently` and `flattenSequentially`. It would be nice to be able to specify the amount of concurrency desired.

Here's a half-baked snippet that I call `flattenParallel`, which handles a variable number (`n`) of streams concurrently:

```javascript
function flattenParallel(n) {
  return function(input$$) {
    const pending = [];
    let active = 0;
    return xs.create({
      start: out => {
        function onNext(input$) {
          if (active > n) pending.push(input$);
          else {
            active++;
            input$.addListener({
              next: item => out.next(item),
              complete: () => {
                active--;
                if (pending.length > 0) onNext(pending.shift());
              }
            });
          }
        }
        input$$.addListener({
          next: onNext
        });
      },
      stop: () => {}
    });
  };
}
```

This provides a flexible abstraction of `flattenSequentially` and `flattenConcurrently`:

- `flattenSequentially` is `flattenParallel(0)`
- `flattenConcurrently` is `flattenParallel(Infinity)`

kloy:
This is the #1 feature keeping me off stream.

staltz:
Apart from "it would be nice", have any of you encountered a real need for this operator when writing real code?

plmrry:
Can't it be used in any situation when you want to be able to specify concurrency?

If you want to read 10,000 files, and you want to open them in parallel, say 20 at a time.

I wrote it to handle bulk API requests – I needed to make thousands of API calls, with some concurrency, but without hitting my rate limit.

staltz:
> For example: If you want to read 10,000 files, and you want to open them in parallel, say 20 at a time.

I'd say that's a use case for Node Streams.

> I wrote it to handle bulk API requests – I needed to make thousands of API calls, with some concurrency, but without hitting my rate limit.

From a client? Sounds like the job of an intermediate backend.

I'm getting to a point here, which is: xstream is meant for frontend programming, and in specific, to support Cycle.js. I find it hard to find a usage for flattenParallel in that context.

plmrry:
I didn't realize the project had such a narrow focus. 

Is there any reason xstream cannot or should not be used in a backend?

kloy:
@staltz Thank you for the replies.

Here are my use cases, which are all from the browser client.

[...]

PR #260, opened 2018-07-24
feat(flattenConcurrentlyAtMost): add new extra

xtianjohns:
Add flattenSequentiallyAtMost(n) extra.

### Description
Add new extra operator file and unit tests. The operator extra started as a copy of `flattenConcurrently` wrapped in a factory function. From there, a few renames and some combined functionality from `flattenSequentially` to implement the inner buffer/queue named `_seq`.

There are some distinctions between both, obviously. The completion method (`_c`) needs to be able to determine if there are any currently connected streams, or any streams in the buffer. If the answer to both questions is no, then we send a completion event downstream.

Additionally, there may be a case when the last inner stream completes, there's no items left in the queue, and the meta stream has completed. This also means we should send a completion event downstream.

To track these fields, the operators gets some new properties, `_l` which tracks how many streams the operator is listening to, `_seq` which represents the buffer and `_d` which indicates whether the meta stream has completed.

Resolves staltz/xstream#161.

### Motivation and Context
flattenSequentiallyAtMost is designed to provide consumer-configurable concurrency to flattening operations. Two flattening extras exist which allow consumers to flatten a meta stream with maximum concurrency, or with no concurrency. This new operator supports a concurrency limit, representing the maximum amount of _additional_ streams to connect to during flattening.

### How Has This Been Tested?
I've added 2 new tests which verify that if an inner stream errors that we send an error downstream, and that we don't greedily connect to more than `n` streams at a time.

I've also duplicated both test suites from `flattenSequentially` and `flattenConcurrently`, and substituted those extras with `flattenConcurrentlyAtMost(1)` and `flattenConcurrentlyAtMost(Infinity)` respectively. This may be overkill, but was a powerful tool during development to know I wasn't violating semantics by adding the configurable buffer.

## Checklist:

### DIFF
--- a/src/extra/flattenConcurrentlyAtMost.ts
+++ b/src/extra/flattenConcurrentlyAtMost.ts
@@ -1,0 +1,132 @@
+import {Operator, Stream, OutSender, InternalListener} from '../index';
+
+class FCAMIL<T> implements InternalListener<T>, OutSender<T> {
+  constructor(public out: Stream<T>,
+              private op: FlattenConcAMOperator<T>) {
+  }
+
+  _n(t: T) {
+    this.out._n(t);
+  }
+
+  _e(err: any) {
+    this.out._e(err);
+  }
+
+  _c() {
+    this.op.less();
+  }
+}
+
+export class FlattenConcAMOperator<T> implements Operator<Stream<T>, T> {
+  public type = 'flattenConcurrentlyAtMost';
+  public out: Stream<T> = null as any;
+  private _l: number = 0;
+  private _d: boolean = false;
+  private _seq: Array<Stream<T>> = [];
+
+  constructor(public n: number, public ins: Stream<Stream<T>>) {
+  }
+
+  _start(out: Stream<T>): void {
+    this.out = out;
+    this.ins._add(this);
+  }
+
+  _stop(): void {
+    this.ins._remove(this);
+    this._l = 0;
+    this.out = null as any;
+    this._seq = [];
+  }
+
+  less(): void {
+    const seq = this._seq;
+    if (--this._l === 0 && seq.length === 0 && this._d) {
+      const u = this.out;
+      if (!u) return;
+      u._c();
+    }
+    if (this._l < this.n && seq.length > 0) {
+      this._n(seq.shift() as Stream<T>);
+    }
+  }
+
+  _n(s: Stream<T>) {
+    const u = this.out;
+    if (!u) return;
+    if (this._l < this.n) {
+      this._l++;
+      s._add(new FCAMIL(u, this));
+    } else {
+      this._seq.push(s);
+    }
+  }
+
+  _e(err: any) {
+    const u = this.out;
+    if (!u) return;
+    u._e(err);
+  }
+
+  _c() {
+    const seq = this._seq;
+    this._d = true;
+    if (this._l === 0 && seq.length === 0) {
+      const u = this.out;
+      if (!u) return;
+      u._c();
+    }
+  }
+}
+
+/**
+ * Flattens a "stream of streams", handling multiple concurrent nested streams
+ * simultaneously, up to some limit `n`.
+ *
+ * If the input stream is a stream that emits streams, then this operator will
+ * return an output stream which is a flat stream: emits regular events. The
+ * flattening happens concurrently, up to the configured limit. It works like
+ * this: when the input stream emits a nested stream,
+ * *flattenConcurrentlyAtMost* will start imitating that nested one. When the
+ * next nested stream is emitted on the input stream,
+ * *flattenConcurrentlyAtMost* will check to see how many streams it is connected
+ * to. If it is connected to a number of streams less than the limit, it will also
+ * imitate that new one, but will continue to imitate the previous nested streams
+ * as well.
+ *
+ * If the limit has already been reached, *flattenConcurrentlyAtMost* will put the
+ * stream in a queue. When any of the streams it is listening to completes, a stream
+ * is taken out of the queue and `flattenConcurrentlyAtMost` will connect to it.
+ *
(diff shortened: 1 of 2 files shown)


src/extra/flattenConcurrentlyAtMost.ts
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  less(): void {
+    const seq = this._seq;
+    if (--this._l === 0 && seq.length === 0 && this._d) {
+      const u = this.out;
+      if (!u) return;
+      u._c();
+    }
+    if (this._l === 0 && seq.length > 0) {
```
staltz: Why do we need this `this._l === 0` check? Suppose `n === 3` and `this._l === 3`, once this `less()` is called, now `this._l === 2` and `2 < 3` so we should try to get stuff from the seq.

xtianjohns: Good catch, that's a bug. We don't want to exhaust the list before draining the queue, I'll fix.

xtianjohns:
Thanks @staltz! I dunno what happened with the message (I got the branch name right), I've rebased the change and I've also grep'd the diff for the name and made sure I'm using the right language.

I've updated the limit to reflect the max for number of streams to listen to, so now `flattenSequentially` and `flattenConcurrentlyAtMost(1)` are equivalent.

staltz:
Thanks @xtianjohns! I noticed a few other issues, mostly with docs/tests. Comments inline

src/extra/flattenConcurrentlyAtMost.ts
```diff
@@ -0,0 +1,132 @@
...
+ *
+ * Marble diagrams:
+ *
+ * ```text
+ * --+--------+---------------
+ *   \        \
+ *    \       ----1----2---3--|
+ *    --a--b----c----|
+ *     flattenConcurrentlyAtMost(1)
+ * -----a--b----c------2---3--|
```
staltz: The output here *could* be correct, as well as it could have been different, e.g.

```
-----a--b----c---------1----2---3--|
```

In case there was no third-party subscriber for the 1-2-3 inner stream. I think this diagram should be the one I just showed because flattenConcurrentlyAtMost(1) is supposed to be like flattenSequentially, and if you see the diagram for that one, it's a-b-c-1-2-3.

src/extra/flattenConcurrentlyAtMost.ts
```diff
@@ -0,0 +1,132 @@
...
+ * ```
+ *
+ * ```text
+ * --+---+---+-|
+ *    \   \   \
+ *     \   \   ---fgh----i-----jh--|
+ *      \   -----1----2----3--|
+ *       ---a--b-----c--|
+ *     flattenConcurrentlyAtMost(2)
+ * ---------a--b-1---c2--i-3---jh--|
```
staltz: Like the previous diagram, I think this one should be 

```
--+---+---+-|
   \   \   \
    \   \   ---fgh----i-----jh--|
     \   -----1----2----3--|
      ---a--b-----c--|
    flattenConcurrentlyAtMost(2)
---------a--b-1---c2--i-3------fgh----i-----jh--|
```

tests/extra/flattenConcurrentlyAtMost.ts
```diff
@@ -0,0 +1,387 @@
+/// <reference types="mocha"/>
+/// <reference types="node" />
+import xs, {Stream, Listener} from '../../src/index';
+import flattenConcurrentlyAtMost from '../../src/extra/flattenConcurrentlyAtMost';
+import * as assert from 'assert';
+
+describe('flattenConcurrentlyAtMost (extra)', () => {
+
+  describe('with n less than Infinity', () => {
+    it('should process only n + 1 streams at a time', (done: any) => {
```
staltz: I think this should be `only n streams`

xtianjohns:
Yep, these make sense @staltz, will fix.

tests/extra/flattenConcurrentlyAtMost.ts
```diff
@@ -0,0 +1,387 @@
...
+
+      it('should expand 3 sync events as a periodic each', (done: any) => {
+        const stream = xs.of(0, 1, 2)
+          .map(i => xs.periodic(100 * (i + 1) + 10 * i).take(2).map(x => `${i}${x}`))
+          .compose(flattenConcurrentlyAtMost(Infinity));
+        // ---x---x---x---x---x---x
+        // ---00--01
+        // --------10------11
+        // -----------20----------21
+        const expected = ['00', '01', '10', '20', '11', '21'];
```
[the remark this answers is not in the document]
xtianjohns: Good catch, I've added this exact case: that the operator subscribes to new streams from its buffer only when it needs new events.

tests/extra/flattenConcurrentlyAtMost.ts
xtianjohns: Agree! Fixed.

src/extra/flattenConcurrentlyAtMost.ts
xtianjohns: Fixed!

src/extra/flattenConcurrentlyAtMost.ts
xtianjohns: Right, good thinking. Makes sense to leave the diagrams as simple as possible, and not assuming other subscribers seems like a good approached. Fixed!

xtianjohns:
Sorry for the wait, @staltz. You may be even _less_ likely to merge now since it's been so long, but hey - at least you can decline a complete PR. 😄

staltz:
@xtianjohns Thanks for the reminder! No reason to decline this PR, I probably just lost it in the middle of many other notifications. :)

staltz:
Released 11.9.0

PR #266, opened 2018-09-24
Fix styling of website

k15a:
Currently the stylesheet is not loading when you visit the website with https as it's trying to load the stylesheet from http. Also, GitHub does not allow do use raw files directly. AFAIK this should fix both.


staltz:
Thanks!

Issue #262, opened 2018-08-31
Can't flatten stream of MemoryStream in TypeScript


zheeeng:
```typescript
declare const ss: Stream<MemoryStream<string>>
declare const ss2: MemoryStream<MemoryStream<string>>
declare const ss3: Stream<Stream<string>>
declare const ss4: MemoryStream<Stream<string>>

ss.flatten()
ss2.flatten()
ss3.flatten()
ss4.flatten()
```

`ss` reports:
```
[ts]
The 'this' context of type 'Stream<MemoryStream<string>>' is not assignable to method's 'this' of type 'Stream<Stream<string>>'.
  Types of property '_ils' are incompatible.
    Type 'InternalListener<MemoryStream<string>>[]' is not assignable to type 'InternalListener<Stream<string>>[]'.
      Type 'InternalListener<MemoryStream<string>>' is not assignable to type 'InternalListener<Stream<string>>'.
        Type 'Stream<string>' is not assignable to type 'MemoryStream<string>'.
```

`ss2` reports:
```
[ts]
The 'this' context of type 'MemoryStream<MemoryStream<string>>' is not assignable to method's 'this' of type 'Stream<Stream<{}>>'.
  Types of property '_prod' are incompatible.
    Type 'InternalProducer<MemoryStream<string>>' is not assignable to type 'InternalProducer<Stream<{}>>'.
      Type 'MemoryStream<string>' is not assignable to type 'Stream<{}>'.
        Types of property '_ils' are incompatible.
          Type 'InternalListener<string>[]' is not assignable to type 'InternalListener<{}>[]'.
            Type 'InternalListener<string>' is not assignable to type 'InternalListener<{}>'.
              Type '{}' is not assignable to type 'string'.
```

Sinewyk:
You can use `"strictFunctionTypes": false` in `tsconfig.json` until this is resolved

atomrc:
I get this error, as well, with xstream 11.11.0 and typescript 3.8.3. 
I am using `@cycle/http` and just do `http.select(cat).flatten()` which gives me

```
The 'this' context of type 'Stream<MemoryStream<Response> & ResponseStream>' is not assignable to method's 'this' of type 'Stream<MemoryStream<Response> | Stream<Response>>'.
  Types of property '_ils' are incompatible.
    Type 'InternalListener<MemoryStream<Response> & ResponseStream>[]' is not assignable to type 'InternalListener<MemoryStream<Response> | Stream<Response>>[]'.
[...]

Issue #231, opened 2017-11-10
Update to TypeScript 2.6.1 with strictFunctionTypes


whitecolor:
[TypeScript 2.6.1 with strictFunctionTypes](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-6.html)

This causes multiple issues in xstream codebase itself, and also it affects typings of different operators (like `flatten`) that can/should be improved/fixed.

mightyiam:
Related to #230.

woutervh-:
Wrote a fix #233

webmaster128:
As I wrote in #233, I think this library can now be used from applications that set strictFunctionTypes = true. So we should close this or clarify that it is only about the test suite.

Sinewyk:
@staltz @whitecolor closed via https://github.com/staltz/xstream/pull/268 or redefine the issue related to types improvement.

PR #268, opened 2018-12-22
 refactor(compile): Activate typescript strict features

Sinewyk:
Just a cleanup in preparation of trying to fix https://github.com/staltz/xstream/issues/262

Actual line that we need to make compile after: https://github.com/Sinewyk/xstream/blob/2299e8f3441ed3b0723e830cfd2000b684858faa/tests/operator/flatten.ts#L338-L341

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1764,8 +1764,8 @@ export class Stream<T> implements InternalListener<T> {
    *
    * @return {Stream}
    */
-  flatten<R>(this: Stream<Stream<R>>): T {
-    return new Stream<R>(new Flatten(this)) as T & Stream<R>;
+  flatten<R>(this: Stream<Stream<R>>): Stream<R> {
+    return new Stream<R>(new Flatten(this));
   }
 
   /**
@@ -1976,7 +1976,7 @@ export class Stream<T> implements InternalListener<T> {
 
 export class MemoryStream<T> extends Stream<T> {
   private _v?: T;
-  private _has: boolean = false;
+  private _has?: boolean = false;
   constructor(producer: InternalProducer<T>) {
     super(producer);
   }
--- a/src/index.ts
+++ b/src/index.ts
@@ -87,7 +87,7 @@ function internalizeProducer<T>(producer: Producer<T> & Partial<InternalProducer
     il.next = il._n;
     il.error = il._e;
     il.complete = il._c;
-    this.start(il);
+    this.start(il as Listener<T>);
   };
   producer._stop = producer.stop;
 }
@@ -119,7 +119,7 @@ class Observer<T> implements Listener<T> {
 class FromObservable<T> implements InternalProducer<T> {
   public type = 'fromObservable';
   public ins: Observable<T>;
-  public out: Stream<T>;
+  public out?: Stream<T>;
   private active: boolean;
   private _sub: Subscription | undefined;
 
@@ -1765,7 +1765,6 @@ export class Stream<T> implements InternalListener<T> {
    * @return {Stream}
    */
   flatten<R>(this: Stream<Stream<R>>): T {
-    const p = this._prod;
     return new Stream<R>(new Flatten(this)) as T & Stream<R>;
   }
 
@@ -1976,7 +1975,7 @@ export class Stream<T> implements InternalListener<T> {
 }
 
 export class MemoryStream<T> extends Stream<T> {
-  private _v: T;
+  private _v?: T;
   private _has: boolean = false;
   constructor(producer: InternalProducer<T>) {
     super(producer);
@@ -1994,14 +1993,14 @@ export class MemoryStream<T> extends Stream<T> {
     const a = this._ils;
     a.push(il);
     if (a.length > 1) {
-      if (this._has) il._n(this._v);
+      if (this._has) il._n(this._v!);
       return;
     }
     if (this._stopID !== NO) {
-      if (this._has) il._n(this._v);
+      if (this._has) il._n(this._v!);
       clearTimeout(this._stopID);
       this._stopID = NO;
-    } else if (this._has) il._n(this._v); else {
+    } else if (this._has) il._n(this._v!); else {
       const p = this._prod;
       if (p !== NO) p._start(this);
     }
--- a/tests/factory/from.ts
+++ b/tests/factory/from.ts
@@ -1,3 +1,5 @@
+/// <reference types="mocha"/>
+/// <reference types="node" />
 import {Promise} from 'es6-promise';
 import xs, { Observable } from '../../src/index';
 import * as assert from 'assert';
@@ -25,7 +27,7 @@ describe('xs.from', () => {
     const stream = xs.from(Promise.reject('no'));
 
     stream.addListener({
-      next: () => done('next should not be called'),
+      next: (x: string) => done('next should not be called'),
       error: (err: any) => {
         assert.strictEqual(err, 'no');
         done();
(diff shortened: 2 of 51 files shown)


staltz:
Thanks @Sinewyk, this looks like a carefully written PR. I have two comments, I'll add them inline.

Also does @jvanbruegge have comments?

src/index.ts
```diff
@@ -1764,9 +1764,8 @@ export class Stream<T> implements InternalListener<T> {
    *
    * @return {Stream}
    */
-  flatten<R>(this: Stream<Stream<R>>): T {
-    const p = this._prod;
-    return new Stream<R>(new Flatten(this)) as T & Stream<R>;
+  flatten<R>(this: Stream<Stream<R>>): Stream<R> {
```
staltz: I wonder what was the need for this change. I remember when I authored this line, that `T` was necessary, `Stream<R>` didn't suffice because `R` is a free variable (often the usage is `num$$.flatten()` not `num$$.flatten<number>()`), but it might be that it just works nowadays with the newest TS. Did you verify that this line *had* to be changed to satisfy compilation?

tests/factory/from.ts
```diff
@@ -24,10 +22,10 @@ describe('xs.from', () => {
   });
 
   it('should convert a rejected promise to a stream', (done: any) => {
-    const stream = xs.from(Promise.reject('no'));
+    const stream = xs.from<string>(Promise.reject('no'));
```
staltz: Is this change (and similar changes like this) necessary because strict features were turned on, or because the `src` changed? I'd like the previous test code to still pass, because that indicates no breaking changes for people who were using it as `xs.from(promise)` not `xs.from<string>(promise)`.

Sinewyk: Haha, actually this one no, I must have gone through the error list, it first errored later because x was unused line 30, I removed it, so there's no need for the type hint anymore =D.

src/index.ts
Sinewyk: I tried to get correct typehints here https://github.com/staltz/xstream/pull/268/files#diff-20c02845523c7301982ad63461750c4fR338.

When looking at the implementation of flatten, it always output a Stream, so T is plain wrong if `Stream<MemoryStream<A>>.flatten` (T being `MemoryStream<A>`) gives `MemoryStream<A>`. It should always give Stream<A>, whatever the combinations of Stream/MemoryStream. It didn't without this change.

Sinewyk:
One thing I think for sure is extraneous in this PR, is all the `/// <reference types="node" />` comment I deleted. Correct me if I'm wrong, but AFAIK they are old, for when typescript did not know how to find them by itself, so they aren't necessary anymore.

But I did not _need_ to delete them to make everything compile.

tests/factory/from.ts
Sinewyk: Went through the whole PR, this was the only case, where fixing one errors fixed 2.

src/index.ts
```diff
@@ -1976,7 +1976,7 @@ export class Stream<T> implements InternalListener<T> {
 
 export class MemoryStream<T> extends Stream<T> {
   private _v?: T;
-  private _has: boolean = false;
+  private _has?: boolean = false;
```
Sinewyk: Not necessary anymore too

src/index.ts
jvanbruegge: @staltz typescript infers those

Sinewyk:
I guess I'll just squash the latest commit then, even though it's green on my machine.

src/index.ts
Sinewyk: Well, it seems it still is.

webmaster128:
Great stuff! This should also close https://github.com/staltz/xstream/pull/233, right?

Sinewyk:
Oh, and I could improve the typing of flatten, by reading #233 I saw `flatten<R, S extends Stream<R>>(this: Stream<S>): Stream<R>`, that was something I was looking for but did not know how to express at the time.

PS: also closes #231 ?

Sinewyk:
Playing around with https://github.com/Sinewyk/xstream/commit/ef27baccac41bc84bd1763badccd9a200d83436b (this branch + another commit, didn't want to pollute this PR with some experiments) to "improve" the flatten operator, around the idea:
```ts
flatten<R, S extends Stream<R>>
```

It works except the `tests/operator/flatten.ts`, I have tons of
```
The 'this' context of type 'Stream<Stream<number>>'
is not assignable to method's 'this' of type 'Stream<Stream<{}>>
```
so I can't quite get the type to be inferred by typescript correctly.

image

Only the third line ends up "good"
```ts
(method) Stream<Stream<number>>.flatten<number, Stream<number>>(this: Stream<Stream<number>>): Stream<number>
```

The second line (the most important because of cyclejs `HTTP.select().flatten()`) is
```ts
(method) Stream<MemoryStream<number>>.flatten<{}, Stream<{}>>(this: Stream<Stream<{}>>): Stream<{}>
```
we can make it work by supplying type parameters, but I'm not satisfied (and it would be a breaking change ?), any ideas ?

staltz:
@Sinewyk It's a good idea to keep PRs focused on solving one thing at a time, e.g. I'm happy to review other PRs if you have other fixes here and there. For now, it's important to do only what the commit message is saying: 'activate typescript strict features'. So, for instance, activate it in tsconfig.json, and do the necessary changes in the source file. The tests files should not change, *unless* they simply do not pass, which means that there is a breaking change. I think this PR would probably benefit from having a tsconfig.json separately for the tests, and the necessary changes to the npm script so that it uses the new tsconfig.json. By using a **strict** tsconfig for the **implementation**, and a **loose** tsconfig for **tests**, we can verify that xstream TypeScript is meant *not only for users that setup strict features*, but also for people who use the loose mode in TypeScript. JavaScript use is the loosest of all, but we don't need to test that in compile-time, obviously, just runtime.

jvanbruegge:
@staltz strict mode is a subset of loose mode. When it compiles in strict modr, it will also compile in loose mode

staltz:
@jvanbruegge Yes I understand that for the implementation, but strict mode was not being applied to tests before (which should reflect not the implementation, but the expected and supported use cases). This is not just about correctness, it's also about what the commit message communicates.

jvanbruegge:
But isnt that the goal of this PR? Making `HTTP.select().flatten()` work with xstream, aka usage, not implementation?

staltz:
Yes, it can be one of the goals, as long as it's done explicitly. For instance, if the goal is to support that API with xstream, then we should *add* a new test that reflects that use case, instead of changing all of the tests in a way that implicitly or not clearly indicates support for this specific use case.

Sinewyk:
Do we need to have two tsconfig.json then ? One for the source code in strict, one for the tests in loose ?

staltz:
@Sinewyk, if possible, yes

Sinewyk:
I'm not happy with this, I don't know why yet.

Random: I think `xstream` should be pulled into the `cyclejs` monorepo, it was created for it, it should live with it.

mightyiam:
I have used xstream without Cycle.js before.

Sinewyk:
Sure, go for it. Here's to hoping I didn't mess something up :beers:

staltz:
Alright, thanks a lot for the PR and for the patience! Well done :)

> random: I use the refined github extension (firefox at least) that move some things around, so it's a lot more difficult (impossible?) to close issues/PR by mistake

Just in case you're wondering, I did that to manually trigger a Travis build.

PR #270, opened 2019-02-15
Add combine typed signature

whitecolor:
Add `combine` method typed signature. Allows getting a typed result when combining a list of homogeneous streams.

```
const item$s = items.map(item => xs.of(item))

const items$ = xs.combine(...items) // <= Stream<Items[]>
```


staltz:
Really useful patch! Thanks!

PR #278, opened 2019-05-02
fix typings for later TS versions (3.4)

whitecolor:
const mm$ = xs.createWithMemory<MemoryStream<number>>()
  const sm$ = xs.create<MemoryStream<number>>()
  const ms$ = xs.createWithMemory<Stream<number>>()
  const ss$ = xs.create<Stream<number>>()
  const fmm$ = mm$.flatten() // without fix this is typed to Stream<{}> not Stream<number>
  const fsm$ = sm$.flatten()
  const fms$ = ms$.flatten()
  const fss$ = ss$.flatten()


staltz:
Thanks!

staltz:
Released 11.11.0

PR #288, opened 2019-12-19
Update URL for dox

ryota-ka:
While working on #287, I've noticed that [visionmedia/dox](https://github.com/visionmedia/dox) is now [tj/dox](https://github.com/tj/dox) (maybe transferred?), causing integration check failures on `yarn install`.

I've updated URL for the tarball specified in `yarn.lock`.


staltz:
Thanks @ryota-ka !

PR #295, opened 2020-01-17
Use `never` for never stream instead of `any`

CYBAI:
Using `never` type should allow us to have the _next_ value of `never` stream with `never` type.


staltz:
Hi, could you please take a look at #287 ?

CYBAI:
oops, sorry, I didn't notice that >_< let me close this one then, thanks!

PR #296, opened 2020-01-28
debounce: clear emit timer interval on stop

whitecolor:
There is a possible bug when timer is set (not yet executed) and stream stopped. If timer interval is not cleared it results in infinite emitting (not sure why emitting is looped).

 I don't have a clean simple repro example, but it happened in my app, had to debug it. 

Anyway it seems logical to clear not executed emit interval on `stop`, and not to clear it is a potential problem.


whitecolor:
@staltz this is a bug.

staltz:
@whitecolor Sorry it took ridiculously long to merge this. I'll publish version 11.12.0

PR #301, opened 2020-04-29
refactor(core): use copy helper instead

jack-cn:


staltz:
Thank you! Sorry it took a while

PR #307, opened 2020-07-22
feat: add never and empty <T = any> param

whitecolor:
This would allow such things without typing issues:
```ts
declare const $some: Stream<number> | undefined

($some || xs.empty<number>()).map(x => -x)
```

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1281,7 +1281,7 @@ export class Stream<T> implements InternalListener<T> {
    * @factory true
    * @return {Stream}
    */
-  static never<T = any>(): Stream<any> {
+  static never<T = any>(): Stream<T> {
     return new Stream<T>({ _start: noop, _stop: noop });
   }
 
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -1281,8 +1281,8 @@ export class Stream<T> implements InternalListener<T> {
    * @factory true
    * @return {Stream}
    */
-  static never(): Stream<any> {
-    return new Stream<any>({ _start: noop, _stop: noop });
+  static never<T = any>(): Stream<any> {
```
staltz, line 1284: Shouldn't this function return `Stream<T>` now?

whitecolor: Yes, my bad.

staltz:
Thanks!

whitecolor:
@staltz will you publish this? seems not breaking)

staltz:
Thanks for the reminder. 11.13.0 is out :+1:

PR #315, opened 2020-09-21
chore(deps): Upgrade symbol-observable to version 2

### LINKED ISSUE #312: Upgrade symbol-observable to ^2
[symbol-observable](https://github.com/benlesh/symbol-observable) version 2 was released, including a [fix](https://github.com/benlesh/symbol-observable/pull/48) that allows us to use [SES](https://github.com/Agoric/SES-shim/tree/master/packages/ses#secure-ecmascript-ses) and [LavaMoat](https://github.com/LavaMoat/lavamoat) as a defense against prototype pollution and supply chain attacks.

The [CHANGELOG](https://github.com/benlesh/symbol-observable/blob/master/CHANGELOG.md) contains the following warning

> BREAKING CHANGE: This moves to using Symbol.for to create the symbol instance. It's possible that some runtimes that support Symbol do not support Symbol.for. Therefor, I'm marking this as a breaking change. If you find that you hit this issue, please report it. You can work around it by polyfilling Symbol.for.

According to caniuse.com, [Symbol](https://www.caniuse.com/mdn-javascript_builtins_symbol) and [Symbol.for](https://www.caniuse.com/mdn-javascript_builtins_symbol_for) have practically the same support. The only difference I see is in Chrome 38-39 (released in 2014) and Opera 25-26 (released in 2014).

kriskowal:
This upgrade brings xstream into the set of npm packages that can be safely run without any mutation of primordial prototypes. Such packages can be used in applications that freeze the prototypes to mitigate prototype pollution supply chain attacks.  For example, `xstream` is in the supply chain for the CosmJS financial instruments project and would benefit from such safety measures.

This is considered a breaking change on account of the unlikely possibility that a platform exists that does not support both Symbol and Symbol.for. This hypothetical platform would no longer be supported.

Closes #312

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,9 @@
   },
   "homepage": "https://github.com/staltz/xstream#readme",
   "dependencies": {
-    "symbol-observable": "1.2.0"
+    "@types/globalthis": "^1.0.1",
+    "globalthis": "^1.0.1",
+    "symbol-observable": "^2.0.3"
   },
   "devDependencies": {
     "@types/mocha": "^2.2.40",
@@ -64,13 +66,13 @@
     "markdox": "0.1.10",
     "mkdirp": "0.5.1",
     "mocha": "^5.2.0",
-    "most": "1.0.3",
+    "most": "^1.9.0",
     "sinon": "1.16.0",
     "strip-comments": "0.4.4",
     "ts-node": "6.0.x",
     "tsify": "4.0.x",
     "tslint": "5.7.0",
-    "typescript": "3.2.x",
+    "typescript": "^3.4.0",
     "validate-commit-msg": "2.4.x"
   },
   "publishConfig": {
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,4 +1,7 @@
-import $$observable from 'symbol-observable';
+import ponyfillSymbolObservable from 'symbol-observable/ponyfill';
+import { getPolyfill as getGlobalThis } from 'globalthis';
+
+const $$observable = ponyfillSymbolObservable(getGlobalThis());
 
 const NO = {};
 function noop() { }
@@ -1415,7 +1418,7 @@ export class Stream<T> implements InternalListener<T> {
    * @return {Stream}
    */
   static fromObservable<T>(obs: { subscribe: any }): Stream<T> {
-    if ((obs as Stream<T>).endWhen) return obs as Stream<T>;
+    if ((obs as Stream<T>).endWhen !== undefined) return obs as Stream<T>;
     const o = typeof obs[$$observable] === 'function' ? obs[$$observable]() : obs;
     return new Stream<T>(new FromObservable(o));
   }
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,15 +2,21 @@
 # yarn lockfile v1
 
 
-"@most/multicast@^1.2.3":
+"@most/multicast@^1.2.5":
   version "1.3.0"
   resolved "https://registry.yarnpkg.com/@most/multicast/-/multicast-1.3.0.tgz#e01574840df634478ac3fabd164c6e830fb3b966"
+  integrity sha512-DWH8AShgp5bXn+auGzf5tzPxvpmEvQJd0CNsApOci1LDF4eAEcnw4HQOr2Jaa+L92NbDYFKBSXxll+i7r1ikvw==
   dependencies:
     "@most/prelude" "^1.4.0"
 
 "@most/prelude@^1.4.0":
   version "1.7.0"
   resolved "https://registry.yarnpkg.com/@most/prelude/-/prelude-1.7.0.tgz#0956ed464ad03e7fc95143eac0c6dd028498d975"
+
+"@types/globalthis@^1.0.1":
+  version "1.0.1"
+  resolved "https://registry.yarnpkg.com/@types/globalthis/-/globalthis-1.0.1.tgz#5310a812c2ef3905b821a1b219ae0e0d12d0e5f1"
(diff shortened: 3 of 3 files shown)


mightyiam:
symbol-observable instead of symbol-observer in the commit message.

kriskowal:
> symbol-observable instead of symbol-observer in the commit message.

Done.

kriskowal:
Ah, I missed an important detail. `xstream` would need to use the ponyfill instead of the polyfill. Working.

kriskowal:
Well, evidently only being passingly familiar with TypeScript has proven a hindrance in converting `xstream` to use the ponyfill instead of the polyfill. It feels like we need a minor version bump on `symbol-observable` to add more type information, specifically about the polyfill module, and maybe even some work to reveal the polyfill at the root of the package instead of under `es` or `lib`. I’ve yet to figure out where to put the types to make it work.

staltz:
Hey @kriskowal, thanks for the PR. It looks all good to me, I don't understand what you meant with ponyfill instead of polyfill, it seems like [symbol-observable is already a ponyfill](https://github.com/benlesh/symbol-observable#symbol-observable-). And I think I would call this a non-breaking change because xstream was released in 2016, two years after the [browsers you mentioned here](https://github.com/staltz/xstream/issues/312) were updated.

The only change I would make is fixing the commit message to not have an exclamation mark, it's not a convention in this repo. But I can do this as a "squash and merge".

kriskowal:
@staltz `symbol-observable`’s documentation is a bit misleading in this regard. It may have drifted from its original form. There are two entry points: `index.js` and `ponyfill.js`. The `index.js` modifies the global environment *and* returns the symbol. The `ponyfill.js` creates and memoizes the symbol. By definition, the “ponyfill” shouldn’t alter the environment under any circumstances, but in practice, it just tolerates a frozen environment by catching an error. This is not as lucid as would be ideal, but for the purposes of compatibility with a frozen environment, the ponyfill entry point should be sufficient. I’m still trying to make this work. I have so far verified that importing `symbol-observable` with a locked down environment does cause an exception during load.

staltz:
@kriskowal Alright, but can you help me understand why change xstream from `index.js` to `ponyfill.js`? Is it so that `symbol-observable`'s `index.js` is not "safe" while `ponyfill.js` is?

kriskowal:
@staltz That is correct. To be fully pedantic, `ponyfill.js` can be used in an environment that is not vulnerable to prototype pollution attacks, so it is the environment that is safe and the ponyfill cooperates. Whereas, running `index.js` in a safe environment throws an exception.

kriskowal:
This is ready for review again. What’s changed?

- Addressed feedback about the commit message.
- `symbol-observable` is up to 2.0.3 with all the necessary TypeScript definitions exported properly.
- `most` is upgraded to 1.9.0, which in turn depends on `symbol-observable` and required a similar change.
- `globalthis` added as a dependency to satisfy the contract of `symbol-observable/ponyfill` with a `globalThis` shim.
- `tsc` upgraded and necessary changes applied to get tests passing, particularly null check made explicitly boolean.

kriskowal:
@staltz Thank you for your feedback. I’ve addressed that and upgraded the necessary dependencies. I’ve not made a change that freezes the the primordials before running tests, but can do that in a follow-up if that’s desirable. For now, the posted changes are sufficient for anyone using `xstream` with frozen primordials.

staltz:
Thanks @kriskowal. One question about the `globalThis` shim: I'm not in full support of that TC39 proposal, but I'd like to know if it's supported in SES and if it's the only way how SES could use the root object? If there's another way, I'd prefer to not have the `globalthis` dependency.

kriskowal:
Interesting. `globalThis` is stage 4 (finished) and rolled out to every browser except IE. Regardless, the `globalthis` shim captures the concern of getting the global from whatever environment you are in, even in `no-eval` CSP environments. Otherwise, the most reliable way to get the global object is `new Function("return this")()`, but that does not work under a `no-eval` CSP.

Otherwise, some JS environments, particularly the XS engine, only realizes the global object by the name `globalThis`.

SES does not require that you use the `globalthis` package. The `globalthis` package just conveniently produces the global object regardless of environment. It could effectively be inlined, though the `globalthis` package goes through the trouble of ensuring that Node.js only sees `module.exports = global`, whereas all other environments see a long cascade of `typedef` checks.

staltz:
@kriskowal Okay, bummer that `globalThis` proposal is official and deployed, I'll merge this then. Thanks for the efforts!

package.json
```diff
@@ -46,7 +46,9 @@
   },
   "homepage": "https://github.com/staltz/xstream#readme",
   "dependencies": {
-    "symbol-observable": "1.2.0"
+    "@types/globalthis": "^1.0.1",
```
webmaster128, line 49: Shouldn't this go to dev dependencies? `globalthis` types are not in the puublic API of xstream, so users don't need to install it.

webmaster128: Sorry I'm too late before merge. Did not want to comment without knowing if this dependency stays at all.

staltz: Oh yeah, well noted. I'll do that in a separate commit, now it's too late to review, this has been merged.

staltz:
Released `xstream@11.14.0`

package.json
kriskowal: This is imported in `src/index.ts`. Is it the case that `tsc` inlines the dependency in a prepublish build step?

staltz: `globalthis` is imported in runtime, but `@types/globalthis` is not, and users of the library will be importing the typescript-to-javascript transpiled code, built on prepublish.

kriskowal: Ah, very good. Thanks!

PR #317, opened 2020-10-22
Make strict mode Stream<T> covariant on T

whitecolor:
This PR:
- enforces TS `strict` mode on the whole repo
- makes `Stream<T>` covariant on `T` by modifying inner `_ils`, `_dl` types.


whitecolor:
@staltz hello, what do you say?

PR #332, opened 2022-01-27
fix: add skipLibCheck to fix build issue

ntilwalli:
@staltz this PR is to be released as either v6.6.1 or v6.7. It adds `skipLibCheck: true` to the tsconfig.


ntilwalli:
This fix was invalid

